PR #8, opened 2025-05-11
Feature/ENG-19: User onboarding

Dennull:
## Related Issues

Closes ENG-19

## Summary

Adds an onboarding screen, which is the first thing users see when they initially download the app. It features a carousel to promote some of our main features. Users can click the "Create an account" button to be redirected to the account creation screen, and the "I already have an account" button to be redirected to the login screen

## Changes Made

The changes seem quite chunky, but I think that's just because we're getting started so there was a lot of stuff to set up and boilerplate to remove. The changes can be pretty much boiled down to these:

- Set up the basic app routing structure (and hence the file directory structure)
  - Users land on the "auth" related routes first, and at the end of that flow they will be redirected to the "protected" routes
  - Also wrapped app in `SafeAreaProvider` so that we can use `react-safe-area-context`. We should use this library over the default `SafeAreaView` from `react-native`
- Created the onboarding screen. This included a new `Carousel` component, which displays a carousel with pagination and uses `react-reanimated` to add some animations when swiping
- Added the app icon and splash screen

Additionally, there were a few miscellaneous tasks and code quality updates:
- Deleted some boilerplate code we didn't need from the initial set up
- Made a few eslint configurations
  - Import sorting
  - Prop sorting
- Swapped from function syntax to arrow functions. In general I think we should be using named functions where possible as well, so I made sure we were using those wherever we could
  - It should be noted that for routes, we need to export a default component because expo router needs that default export to know what to display on the screen. In these cases I still swapped to using arrow functions but separately exported the default component at the bottom of the file
  - Added and updated some libraries that we need. Check package.json for all of the changes

## Screenshots

https://github.com/user-attachments/assets/3aaf9ec2-a7c6-4593-9c29-b71d1b94ea1e

## Testing Instructions

No additional notes


app/(auth)/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+      className="flex-1 bg-gray-950"
+      resizeMode="cover"
+      source={require("@/assets/images/onboarding-background.png")}
+    >
+      <SafeAreaView className="flex-1 items-center justify-between">
+        <StatusBar barStyle="light-content" />
+        <Carousel data={carouselData} />
+        <View className="flex w-full gap-4 px-8 pb-8">
+          <Pressable className="rounded-lg bg-purple-600 p-3">
+            <Text className="text-center font-clash font-semibold uppercase text-white">
```
NasihNazeem: I noticed a few of the Text components in the files have the font type or the weight included. You should be able to achieve the same result using something like `pbk-h7` if that's the right font size you are looking for in this case — although I think `pbk-h6` is more appropriate to have a bigger button too

Within the same line, I noticed you used `text-white` which is the default white. I mean it's the exact same thing but there's a white in the theme called `text-base-white` as well. The big thing is that we should make it consistent throughout the app incase we need to make changes later. What do you suggest?

components/PressableLink.tsx
NasihNazeem, line 1: Nice component! This makes it easier to handle links moving forward.

assets/images/onboarding-image2@2x.png
NasihNazeem, line 1: I think this is a duplicate of `assets/images/onboarding-image-2@2x.png`

assets/images/onboarding-image2@3x.png
NasihNazeem, line 1: I think this is a duplicate of `assets/images/onboarding-image-2@3x.png`

assets/images/onboarding-image3@2x.png
NasihNazeem, line 1: I think this is a duplicate of `assets/images/onboarding-image-3@2x.png`

assets/images/onboarding-image@2x.png
NasihNazeem, line 1: I think this is a duplicate of `assets/images/onboarding-image-1@2x.png`

assets/images/onboarding-image@3x.png
NasihNazeem, line 1: I think this is a duplicate of `assets/images/onboarding-image-1@3x.png`

components/Carousel.tsx
NasihNazeem, line 1: I'm not all that familiar with the reanimated library so I might have to defer to @JayChuang8 for his thoughts.

components/Carousel.tsx
```diff
@@ -0,0 +1,131 @@
...
+    <View
+      className="flex items-center justify-center gap-4 px-4"
+      style={{ width }}
+    >
+      <Image
+        className="h-[200px] w-[300px]"
+        resizeMode="contain"
+        source={item.image}
+      />
+      <Text className="text-center font-clash text-pbk-h5 font-semibold text-white">
```
NasihNazeem: for text related styling I suggest

`pbk-h5` rather than `font-clash text-pbk-h5 font-semibold`

components/Carousel.tsx
```diff
@@ -0,0 +1,131 @@
...
+        {item.title}
+      </Text>
+      <Text className="text-center font-manrope text-pbk-b1 text-white">
```
NasihNazeem: for text related styling I suggest

`pbk-b1` rather than `font-manrope text-pbk-b1`

components/PressableLink.tsx
```diff
@@ -0,0 +1,26 @@
...
+  href: Href;
+  label: string;
+};
+
+const PressableLink = forwardRef<View, PressableLinkProps>(
+  ({ href, label, ...rest }: PressableLinkProps, ref) => {
+    return (
+      <Link asChild href={href}>
+        <Pressable ref={ref} {...rest}>
+          <Text className="text-center font-clash font-semibold uppercase text-white">
```
NasihNazeem: for font related styling I suggest 

`pbk-h7` instead of `font-clash font-semibold` 

although I also suggest going h6 for the heading size

app/(auth)/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+              Create an Account
+            </Text>
+          </Pressable>
+          <PressableLink
+            className="rounded-lg p-3"
```
NasihNazeem: I noticed that the link is clickable outside of the actual text (like horizontally). I suggest using marginHorizontal: auto or `mx-auto` to deal with this.

NasihNazeem (changes requested):
The carousel looks amazing! The images resolution was something you mentioned in our weekly standup but honestly on my screen it's looking pretty crisp. The one thing I would mention about the images including the background is that we should figure out a way to preload them so that the user has to wait for them to load, even if its only a second or two.

I like the overall folder structure as well. I like that you separated the folder into a protected folder rather than simply calling it something like (app) which is what I initially did. It makes that distinction more clear.

app/(auth)/index.tsx
Dennull: I think for the button text we don't have that in our theme yet, I don't see it on the figma. I'm gonna use `pbk-h7` for now as suggested until we get an update on that. But yes I'll update the rest of the text to use our theme as much as possible too

As for whether `pbk-h6` is better for buttons, do you think we should bring that up with Nasrif?

assets/images/onboarding-image2@2x.png
Dennull: Yes thanks I think I was trying something out and forgot to delete these extra images

app/(auth)/index.tsx
Dennull: Ok sure, if we want to make only the text clickable rather than the whole button I can do that. Ended up adding `self-center` to fix this

NasihNazeem: Cool that works. My bad, I just realized this was supposed to be a button. I can't really see the outline around the background of the button. That's more in-line with the Figma as well, but this is just a nitpick to me. I'm okay with either a link or a button

app/(auth)/index.tsx
NasihNazeem: >  I'm gonna use pbk-h7 for now as suggested until we get an update on that. But yes I'll update the rest of the text to use our theme as much as possible too

Perfect, thanks Dennis

> As for whether pbk-h6 is better for buttons, do you think we should bring that up with Nasrif?

It could be that it's just smaller on my screen but a good size on other screens. I don't think this should be a blocker but I'll bring it up in our next meeting to see if you guys agree about the size of the button on my end

NasihNazeem (approved):
Looks good to me! I'm ready to merge this in.

assets/images/adaptive-icon.png
JayChuang8, line 1: Can this and all the other onboarding images be placed in an `onboarding` folder please? I have the profile and team avatars in their own folders in the images folder as well, so it'll clean things up

Dennull: Yes good idea

app/(auth)/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+  },
+  {
+    title: "Climb the ranked ladder 🏆",
+    image: require("@/assets/images/onboarding-image-6.png"),
+    description:
+      "Compete in the daily matchups, win your games, and climb the ranks to the top!",
+  },
+];
+
+const OnboardingScreen = () => (
```
JayChuang8: Minor nit, can we remove the "Screen" at the end of the component name. To keep it consistent with Nasih and I's components. Same thing for all the other components in this PR

components/Carousel.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const viewabilityConfig = {
+    itemVisiblePercentThreshold: 50,
+  };
+
+  const onViewableItemsChanged = useCallback(
+    ({ viewableItems }: { viewableItems: ViewToken[] }) => {
+      if (
+        viewableItems[0] !== undefined &&
+        viewableItems[0].index !== undefined &&
+        viewableItems[0].index !== null
```
JayChuang8: Should we make an `isNotNil()` function so we can get rid of all the undefined and null checks?

components/Carousel.tsx
```diff
@@ -0,0 +1,127 @@
...
+  return (
+    <View className="flex flex-1 items-center justify-center gap-16">
+      <Animated.FlatList
+        className="flex-none"
+        data={data}
+        horizontal
+        onScroll={onScroll}
+        pagingEnabled
+        renderItem={({ item }) => <CarouselItem item={item} />}
+        showsHorizontalScrollIndicator={false}
```
JayChuang8, line 108: I don't think we need this prop

JayChuang8 (changes requested):
Looks good and functions nicely!

app/(auth)/index.tsx
NasihNazeem: I think I've got a few components with Screen at the end of the component too, I can change those

components/Carousel.tsx
Dennull: This disables the horizontal scroller, which we want. I think you can't see it on iOS but maybe you can see it on Android or web, so I think it can't hurt to include this

components/Carousel.tsx
Dennull: Did you want to make our own custom function or do you want to use lodash?

Dennull: As discussed offline, I'm gonna make our own custom function for now and if we want to use lodash we can install it and change the function later 👍

JayChuang8 (approved):

PR #9, opened 2025-05-15
Feature/ENG-13: Implement user authentication with Firebase and Redux integration

NasihNazeem:
## Related Issues

Closes ENG-13.

## Summary

This PR introduces user authentication using Firebase Authentication, supporting both email/password and Google SSO sign-in. It also integrates Redux Toolkit for user state management and leverages Firebase Functions to automatically create a user document in Firestore when a new user registers, ensuring backend and frontend code are cleanly separated.

## Changes Made

- Integrated Firebase Authentication for email/password and Google SSO sign-in
- Set up Redux and Redux Toolkit, including a user slice and boilerplate for managing user state
- Added Firebase Cloud Function to create a user in the users collection upon new registration
- Separated Firebase functions with the frontend code
- Implemented user flows for login, sign up, and forgot password. 
    - Uses `react-hook-form` and `yup` for form validation
- Created an AuthListener component to sync authentication state with Redux
- Added a Spinner component for loading state

## Screenshots

No additional screenshots

## Testing Instructions

Please use the latest build on EAS as this PR has native libraries that are only accessible on the latest build of EAS.


client/app/(auth)/confirmReset.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  const handleResendEmail = async () => {
+    setIsLoading(true);
+
+    try {
+      const emailToUse = await findEmail(email);
+
+      await sendPasswordResetEmail(auth, emailToUse);
+
+      router.push("/(auth)/confirmReset");
```
JayChuang8: Is this needed here when we're already on the confirm reset page?

client/app/(auth)/confirmReset.tsx
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    return data;
+  };
+
+  const handleResendEmail = async () => {
+    setIsLoading(true);
+
+    try {
+      const emailToUse = await findEmail(email);
```
JayChuang8: I don't think this function is needed here since we only use email on this page. If the email doesn't exist, sendPasswordResetEmail should handle it properly by itself so we don't need to worry about that

client/app/(auth)/confirmReset.tsx
```diff
@@ -0,0 +1,114 @@
...
+              className="min-h-14 w-full items-center justify-center rounded-md bg-purple-600"
+              onPress={() => router.replace("/(auth)")}
+            >
+              <Text className="pbk-h6 text-center text-base-white">
+                RETURN TO LOGIN
+              </Text>
+            </Pressable>
+            <Pressable
+              className="my-8 min-h-14 w-full items-center justify-center rounded-md bg-gray-950"
+              onPress={() => handleResendEmail()}
```
JayChuang8: Is there a downside of allowing the user to spam the resend email button? I know some sites have a 1 minute cooldown or something, does spamming it impact the firebase resources at all?

client/app/(auth)/confirmReset.tsx
JayChuang8, line 1: Pressing the "Resent Email" button the second time in a row seems to result in an error: 

` (NOBRIDGE) ERROR  Login error: [Error: firebase.firestore().collection().where(_, _, *) 'value' argument expected.]`.

 I think it should be fixed when the findEmail function is removed though

client/app/(auth)/index.tsx
```diff
@@ -49,29 +50,35 @@ const carouselData: CarouselData[] = [
...
+    <ImageBackground
+      className="flex-1 bg-gray-950"
+      resizeMode="cover"
+      source={require("@/assets/images/onboarding-background.png")}
+    >
+      <SafeAreaView className="flex-1 items-center justify-between">
+        <StatusBar barStyle="light-content" />
+        <Carousel data={carouselData} />
+        <View className="flex w-full gap-4 px-8 pb-8">
+          <Pressable
```
JayChuang8: Lets use PressableLink here as well

client/app/(auth)/login.tsx
```diff
@@ -0,0 +1,283 @@
...
+    mode: "onSubmit",
+    defaultValues: {
+      email: "",
+      password: "",
+    },
+  });
+
+  const logInWithProvider = async (provider: string) => {
+    try {
+      if (provider === "google") {
```
JayChuang8: Can we store the provider in a `LoginProvider` enum type? And then we can have `LoginProvider.Apple` and `LoginProvider.Google` to use instead of the string

client/app/(auth)/login.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+        if (userData?.createdAt instanceof firestore.Timestamp) {
+          userData.createdAt = userData.createdAt.toDate().toISOString();
+        }
+
+        if (userData?.updatedAt instanceof firestore.Timestamp) {
+          userData.updatedAt = userData.updatedAt.toDate().toISOString();
+        }
+
+        dispatch(setUser(userData as CurrentUser));
```
JayChuang8: I think we should manually map the firestore data into the CurrentUser object rather than casting it. 
`const user = {
  id: user.uid,
  email: userData.email,
  createdAt: userData.createdAt?.toDate(),
  updatedAt: userData.updatedAt?.toDate()
}`
This should also help get rid of the if statements

JayChuang8: On a related note, do we need to store `createdAt` and `updatedAt` in the userSlice? I'm not sure if we'll ever use it in the app, maybe we can just keep it in the db

client/app/(auth)/login.tsx
```diff
@@ -0,0 +1,283 @@
...
+        router.push("/(protected)");
+      } else {
+        dispatch(
+          setUser({
+            userId: user.uid,
+            email: user.email ?? undefined,
+            isLoading: false,
```
JayChuang8: I'm not sure if I understand this, do we need the `isLoading` in the userSlice? I think we should handle it an the component level right?

client/app/(auth)/login.tsx
```diff
@@ -0,0 +1,283 @@
...
+type FormData = {
+  email: string;
+  password: string;
+};
+
+WebBrowser.maybeCompleteAuthSession();
+
+const LoginScreen = () => {
+  const [hidePassword, setHidePassword] = useState(true);
+  const [loading, setLoading] = useState(false);
```
JayChuang8: Minor nit, can we rename it `isLoading` and `setIsLoading`?

client/app/(auth)/login.tsx
```diff
@@ -0,0 +1,283 @@
...
+          setError("password", {
+            type: "manual",
+            message: "Incorrect password.",
+          });
+        } else if (firebaseError.code === "auth/invalid-credential") {
+          setError("root", {
+            type: "manual",
+            message: "These are invalid credentials. Please try again.",
+          });
+        }
```
JayChuang8, line 83: I think we should be less specific about the error when it comes to login for better security. Something similar to this should work:
`if (
  firebaseError.code === "auth/invalid-email" ||
  firebaseError.code === "auth/wrong-password" ||
  firebaseError.code === "auth/invalid-credential" ||
  firebaseError.code === "auth/user-not-found"
) {
  setError("root", {
    type: "manual",
    message: "Invalid email or password. Please try again.",
  });
} else {
  setError("root", {
    type: "manual",
    message: "Something went wrong. Please try again later.",
  });
}`

client/app/(auth)/signUp.tsx
```diff
@@ -0,0 +1,270 @@
...
+import Spinner from "@/components/Spinner";
+import { useAppDispatch } from "@/state/hooks";
+import { setUser } from "@/state/slices/userSlice";
+import { signInWithGoogle } from "@/utils/socialAuth";
+
+const schema = yup.object({
+  email: yup
+    .string()
+    .required("Email is required")
+    .matches(/^[^\s@]+@[^\s@]+\.[^\s@]+$/, "Enter a valid email address"),
```
JayChuang8: Should we also use `.email()` here as well like in `login.tsx`?

client/app/(auth)/signUp.tsx
```diff
@@ -0,0 +1,270 @@
...
+  password: yup
+    .string()
+    .required("Password is required")
+    .min(8, "Password must be at least 88 characters"),
```
JayChuang8: Minor nit, typo in 88

client/components/AuthListener.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { ActivityIndicator, View } from "react-native";
+import { useAppDispatch } from "@/state/hooks";
+import {
+  setUser,
+  CurrentUser,
+  clearUser,
+  setIsLoading,
+} from "@/state/slices/userSlice";
+
+type Props = {
```
JayChuang8: Minor nit, can we rename this to `AuthListenerProps` ([ComponentName]Props)?

client/app/(auth)/signUp.tsx
```diff
@@ -0,0 +1,270 @@
...
+              <Text className="pbk-b3 mb-4 text-red-600">{errorMessage}</Text>
+            )}
+          </View>
+          <Pressable
+            className={`${!isValid ? "bg-purple-900" : "bg-purple-600"} min-h-12 w-full justify-center rounded-md`}
+            disabled={!isValid}
+            onPress={handleSubmit(signUpUser)}
+          >
+            {loading ? (
+              <Spinner />
```
JayChuang8, line 205: Seems like the spinner isn't centered when sign up is pressed

JayChuang8 (changes requested):
Sign up and login works well!

JayChuang8:
Should we add a function where the user has to confirm their email before we create a document in the users collection? Right now someone can sign up with anyone's email even if it isn't theirs

NasihNazeem:
> Should we add a function where the user has to confirm their email before we create a document in the users collection? Right now someone can sign up with anyone's email even if it isn't theirs

Yes good point, will make that adjustment

client/app/(auth)/signUp.tsx
NasihNazeem: Thank you for catching this!

client/app/(auth)/login.tsx
NasihNazeem: Got it, thank you

client/app/(auth)/confirmReset.tsx
NasihNazeem: No I will remove this, thanks

client/app/(auth)/confirmReset.tsx
NasihNazeem: It would affect the quota I believe, so I'll set up a 1-minute cooldown to handle this

client/app/(auth)/login.tsx
NasihNazeem: I'll set up a types file with both Google and Apple (for future use)

client/app/(auth)/login.tsx
NasihNazeem: Got it that makes sense, thank you

client/app/(auth)/index.tsx
NasihNazeem: Nice thanks for catching this

client/app/(auth)/confirmReset.tsx
NasihNazeem: Fixed after removing the findEmail func

client/app/(auth)/signUp.tsx
NasihNazeem: Fixed, thank you

client/app/(auth)/login.tsx
NasihNazeem: We don't really need to, it might just be something that we can pull on request by the user later down the line if a user wants to know. I mapped the rest of the data as requested.

client/app/(auth)/login.tsx
```diff
@@ -0,0 +1,282 @@
...
+  };
+
+  const handleLogin = async (data: LoginFormData) => {
+    setIsLoading(true);
+    try {
+      await signInWithEmailAndPassword(
+        auth,
+        data.email.trim().toLowerCase(),
+        data.password,
+      );
```
JayChuang8, line 62: Do we need `router.replace("/(auth)/createProfile");` here?

client/app/(auth)/signUp.tsx
```diff
@@ -0,0 +1,282 @@
...
+            {errorMessage && (
+              <Text className="pbk-b3 mb-4 text-red-600">{errorMessage}</Text>
+            )}
+          </View>
+          <Pressable
+            className={`${!isValid ? "bg-purple-900" : "bg-purple-600"} min-h-12 w-full items-center justify-center rounded-md`}
+            disabled={!isValid}
+            onPress={handleSubmit(signUpUser)}
+          >
+            {loading ? (
```
JayChuang8: It seems like the spinner stays spinning after the "Email already in use" error is shown. Also minor nit, can we rename this to `isLoading` as well please?

client/app/(auth)/login.tsx
```diff
@@ -0,0 +1,282 @@
...
+  return (
+    <View className="flex-1 bg-gray-950">
+      <TouchableWithoutFeedback onPress={Keyboard.dismiss}>
+        <KeyboardAvoidingView
+          behavior="padding"
+          className="flex-1 flex-col px-6 py-4"
+        >
+          <Pressable
+            className="my-16 size-8 items-center justify-center self-end rounded-md border border-gray-900 p-4"
+            onPress={() => router.back()}
```
JayChuang8: I'm wondering if we should change this to redirect to the onboarding screen rather than going back. If the user gets to the login screen from the reset password screen, pressing the X will bring them back to the reset password screen rather than the onboarding screen. We'd probably have to change the router animation for the onboarding screen though, so it's not a "forward slide" animation.

server/functions/src/index.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const createUserInDatabase = functions.auth
+  .user()
+  .onCreate(async (user: UserRecord) => {
+    const { uid, email, displayName: username, photoURL } = user;
+    const usersRef = admin.firestore().collection("users");
+
+    try {
+      await usersRef.doc(uid).set({
+        userId: uid,
```
JayChuang8: Is it possible to change this to just `id` since we know it's on the user document? Or is it automatically set by firestore?

client/app/(auth)/signUp.tsx
```diff
@@ -0,0 +1,282 @@
...
+  });
+
+  const signUpWithProvider = async (provider: LoginProvider) => {
+    try {
+      if (provider === LoginProvider.Google) {
+        const { user } = await signInWithGoogle();
+        dispatch(
+          setUser({
+            userId: user.uid,
+            email: user.email ?? undefined,
```
Dennull: Is this ever supposed to be undefined? If it's possible, should we throw an error or at least place this in an if check similar to what we do for signing up with username and password on line 92? Actually though, we also do this null coalesce on line 104, so we should also follow whatever we do here there

In my opinion if for some reason we don't get an email back from the server it should be an error, that doesn't make sense

client/app/(auth)/signUp.tsx
```diff
@@ -0,0 +1,282 @@
...
+              userId: user.uid,
+              email: user.email ?? undefined,
+            }),
+          );
+
+          setLoading(false);
+          router.replace("/(auth)/createProfile");
+        }
+      }, 3000);
+    } catch (error) {
```
Dennull, line 72: I'd love to handle errors in a more structured way and abstract it into a function, but honestly that can be a later problem lol

client/app/(auth)/signUp.tsx
```diff
@@ -0,0 +1,282 @@
...
+      const { user } = await createUserWithEmailAndPassword(
+        auth,
+        email,
+        password,
+      );
+      if (user) {
+        await sendEmailVerification(user);
+        console.log("Verification email sent");
+      }
+      const intervalId = setInterval(async () => {
```
Dennull: Can you please explain why we need this interval?

client/app/(auth)/signUp.tsx
```diff
@@ -0,0 +1,282 @@
...
+          type: "manual",
+          message: "Unexpected error occurred.",
+        });
+      }
+    }
+  };
+
+  return (
+    <View className="flex-1 bg-gray-950">
+      <TouchableWithoutFeedback onPress={Keyboard.dismiss}>
```
Dennull: Should we swap to using `Pressable`? It seems to be favored by most people, even the react native docs say so. Or does this not work with `Pressable`?

client/app/(auth)/signUp.tsx
```diff
@@ -0,0 +1,282 @@
...
+        <KeyboardAvoidingView
+          behavior="padding"
+          className="flex-1 flex-col px-6 py-4"
+        >
+          <Pressable
+            className="my-16 size-8 items-center justify-center self-end rounded-md border border-gray-900 p-4"
+            onPress={() => router.back()}
+          >
+            <X color="white" size={20} weight="bold" />
```
Dennull: Do people ever need to return to the carousel? Also, should this be a back button to keep consistent with the rest of the app? It seems like we use back buttons for screen changes, the "X" seems like something for dismissing modals

We should also figure out if we want users to be able to swipe to return back, that's a whole other discussion

client/app/(protected)/_layout.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+const AppLayout = () => {
+  const isSignedIn = useAppSelector(selectCurrentUserId);
+  const isRegistered = useAppSelector(selectIsRegistered);
+
+  if (isSignedIn && !isRegistered) {
+    return <Redirect href="/(auth)/createProfile" />;
+  }
+
+  if (!isSignedIn) {
```
Dennull, line 12: Minor thing, this should go first because if the user isn't signed in we can return earlier. I think it also reads better cause that's what we should check first

client/app/_layout.tsx
```diff
@@ -1,27 +1,35 @@
...
 import { GestureHandlerRootView } from "react-native-gesture-handler";
 import { SafeAreaProvider } from "react-native-safe-area-context";
-import "react-native-reanimated";
+import { Provider } from "react-redux";
+import { store } from "../state/store";
+import AuthListener from "@/components/AuthListener";
 import { GluestackUIProvider } from "@/components/ui/gluestack-ui-provider";
-import "@/global.css";
+
+global.RNFB_SILENCE_MODULAR_DEPRECATION_WARNINGS = true;
```
Dennull, line 15: What is this? If we need it can we place it at the top?

client/utils/cn.ts
```diff
@@ -0,0 +1,6 @@
+import { clsx } from "clsx";
+import { twMerge } from "tailwind-merge";
+
+export function cn(...inputs: any[]) {
```
Dennull: Do you think this belongs in `jsUtils.ts`?

JayChuang8 (changes requested):

client/app/(auth)/signUp.tsx
```diff
@@ -0,0 +1,283 @@
...
+              By signing up, you agree to our
+            </Text>
+            <Pressable onPress={() => router.back()}>
+              <Text className="text-purple-600"> Terms of Service</Text>
+            </Pressable>
+            <Text className="text-gray-600"> and </Text>
+            <Pressable onPress={() => router.back()}>
+              <Text className="text-purple-600">Privacy Policy</Text>
+            </Pressable>
+            <Text className="text-gray-600">.</Text>
```
Dennull: A few things going on here. I think the better way to style this might be to wrap the entire sentence in a text block and then have nested text blocks, similar to p and span tags on web. The markup reads a bit easier and more importantly, the spaces will take into account line breaks if we do this. For example, on my phone there's a weird gap between the beginning of the line and the "and" because my screen size is smaller

The other thing is that we can probably get away with only using a `Text` tag an putting the `onPress` on that for links. Honestly I'm not sure when we should use this over text buttons, maybe it should be a style thing like the text buttons have more padding and specific fonts?

IMG_4413

client/app/(auth)/signUp.tsx
```diff
@@ -0,0 +1,283 @@
...
+          </View>
+
+          <View className="mb-5 flex-row items-center gap-2">
+            <View className="flex-1 bg-gray-800" />
+            <Text className="pbk-b1 text-center text-gray-800">or</Text>
+            <View className="flex-1 bg-gray-800" />
+          </View>
```
Dennull: This doesn't look the same as on the login screen, the line is missing on this screen

client/app/(auth)/signUp.tsx
Dennull, line 1: A few general comments

- The SSO login/signup process looks similar between this and the signup screen (eg. UI and callbacks), do you think we can combine it into a component?
- Should we add something that lets you get to the login screen from this page in case you clicked the wrong thing by accident? For example an "Already have an account?" link. Vice-versa for the login screen

client/app/(auth)/login.tsx
Dennull: Also I forgot, do you guys know what the difference between `router.replace()` and `router.push()` is? And which one we want for these situations?

client/app/(auth)/login.tsx
Dennull: Yeah, I think there needs to be a wider discussion and research done about how we should handle page animations when swapping routes, so maybe we can have a chat about this before making any changes

client/app/(auth)/login.tsx
Dennull, line 1: Just a reminder that some of our comments on the sign up page may also apply here since they're so similar

client/app/(auth)/forgotPassword.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+          <Text className="pbk-b1 mb-8 text-base-white">
+            Enter your email address or username and we will send you a link to
+            reset your password.
+          </Text>
+
+          <Text className="pbk-b2 mb-1.5 text-base-white">
+            Email or username
+          </Text>
+          <Controller
```
Dennull, line 113: Nitpick, but the spacing between this form input and the next button isn't consistent with the other login screens

client/eslint.config.mjs
```diff
@@ -40,7 +40,10 @@ export default defineConfig([
         version: "detect",
       },
       "import/resolver": {
-        typescript: true,
+        typescript: {
+          alwaysTryTypes: true,
+          project: "./tsconfig.json",
```
Dennull, line 45: Do you know why we need this? Is it because of the folder structure change?

client/declarations.d.ts
Dennull, line 1: I honestly never thought about this, but is there a difference between `declarations.d.ts` and `global.d.ts`? Like do you know if they're separate files and what goes in them?

client/components/AuthListener.tsx
Dennull, line 1: For my understanding this fetches for the user who's logged in and displays a loader until we can get that right?

client/components/AuthListener.tsx
```diff
@@ -0,0 +1,62 @@
...
+      setInitializing(false);
+    });
+
+    return subscriber;
+  }, [auth, dispatch]);
+
+  if (initializing) {
+    return (
+      <View style={{ flex: 1, justifyContent: "center", alignItems: "center" }}>
+        <ActivityIndicator size="large" />
```
Dennull: Any reason to choose this over our own custom spinner? Is one better than the other in certain use cases?

client/state/slices/userSlice.ts
```diff
@@ -0,0 +1,31 @@
...
+      state.username = action.payload;
+    },
+    clearUser: () => initialState,
+  },
+});
+
+export const selectCurrentUserId = (state: RootState) => state.user.userId;
+
+export const selectIsUserSignedIn = (state: RootState): boolean =>
+  typeof state.user.userId === "string" && state.user.userId.length > 0;
```
Dennull: Do we need this `typeof` check? Also, if we're checking for whether the string is empty or not I think something like this also works but is simpler

```suggestion
  !!state.user.userId
```

client/state/slices/userSlice.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const selectIsRegistered = (state: RootState): boolean =>
+  state.user.userId !== undefined && state.user.username !== undefined;
```
Dennull: We have a helper function for this called `isNotNil` now

```suggestion
  isNotNil(state.user.userId) && isNotNil(state.user.username)
```

client/utils/socialAuth.ts
Dennull, line 1: What do you think about naming this file `authUtils.ts`? Could potentially move other auth related utility functions in here too

server/functions/.eslintrc.mjs
Dennull, line 1: Was this eslint stuff set up using a cli tool? It's using eslint v8

server/functions/.prettierrc
Dennull, line 1: Should this not be the same as the one in our client folder?

server/functions/src/index.ts
```diff
@@ -0,0 +1,26 @@
...
+        email: email || "",
+        username: username || "",
```
Dennull: Should these be allowed do be empty?

Dennull (changes requested):
I left a lot of comments lol but overall it looks nice and works as it should, so great job!

client/app/(auth)/login.tsx
NasihNazeem: After testing I don't think we need to add a `router.replace()` here. It should be sending a user to `createProfile` here if they have not already registered. If they have, it redirects them to the protected routes. Can you describe or help me recreate the bug you are seeing?

My understanding is if we want the user to be able to go back to the previous screen we use push to maintain a history of pages, whereas replace starts a new 'stack' of pages.
In this case, we don't really want the user to be able to go back after signing up and verifying their email, so I will opt for a `router.replace()`.

client/app/(auth)/login.tsx
NasihNazeem: I think I've found a good solution for this. There's a method called `dismissAll()` that returns to the first screen in the history stack. So it provides a back animation and goes to the onboarding screen.

client/eslint.config.mjs
NasihNazeem: yeah, when running the workflow file on github actions it couldn't find the directory and this seemed to have fixed it.

client/declarations.d.ts
NasihNazeem: Good question, I initially had them in the same file and the global declaration was not being recognized so I created its own file. After a bit more reading, I think the issues comes down to whether I want to tell the compiler to recognize the entire file as a module (`export {}`), or to allow multiple declarations that are not globally applied. 

When applying the global types, you need to use that export statement to apply everything globally, and if I have that statement in a file with the other declarations, those other declarations do not get recognized.

client/utils/socialAuth.ts
NasihNazeem: I like this suggestion, thanks

client/utils/cn.ts
NasihNazeem: Yeah sure thing, it's definitely a genereal utility func that we'll need to use all over the app.

client/state/slices/userSlice.ts
NasihNazeem: Good catch, I just had that typeof check because it shouldn't be undefined, but I think that is covered with the suggested change as a falsy value

client/state/slices/userSlice.ts
NasihNazeem: thanks!

client/app/(auth)/signUp.tsx
NasihNazeem: yes agreed. It got really messy when I was initially handling them. Plus I think there are a few more cases I didn't cover, I stuck to the most likely, and the other cases just have a general message.

client/app/(auth)/signUp.tsx
NasihNazeem: Good catch, I've used this in previous projects but I didn't realize they recommend to use Pressable now.

client/app/(auth)/signUp.tsx
NasihNazeem: > Do people ever need to return to the carousel?

I mean the carousel itself probably not, but the current set up has the carousel as the index.tsx. Meaning whenever they are logged out, that'll be the first thing they see. If we want to change this, how about making the login page the default, and re-route to the onboarding screen if it's a first time user. 

>  Also, should this be a back button to keep consistent with the rest of the app? It seems like we use back buttons for screen changes, the "X" seems like something for dismissing modals

Think this is something we can tag Nasrif on, as I agree that there is some mixed messaging here. I'll implement the back button now as it makes most sense and we can update Nasrif and see what he thinks.

> We should also figure out if we want users to able able to swipe to return back, that's a whole other discussion

Yes I agree, currently it's enabled by default but it would be good to consider why we need it

client/app/(auth)/signUp.tsx
NasihNazeem: Good catch thank you, to fix this I'm moving the setIsLoading call from within the try block to a finally block in `signUpUser()`

client/app/(auth)/signUp.tsx
NasihNazeem: thanks for catching this, missed adding the `h-px`

client/app/(auth)/signUp.tsx
NasihNazeem: Cool thanks for bringing this up. I wasn't sure what was best and turns out there were a few other things wrong with it too. I removed the View component and wrapped everything inside a larger Text component and removed the Pressable's. I replaced them with in-line Text components as suggested with an onPress function. I feel like if this had more styling I would try to use a text button like you mentioned, but this seemed simple enough that I don't think warrants any additional component.

Let me know if you are happy with the changes.

This is what I am going with: 


```suggestion
        <Text className="pbk-b1 my-5 flex-row flex-wrap text-gray-600">
            By signing up, you agree to our{" "}
            <Text className="text-purple-600">Terms of Service</Text> and{" "}
            <Text className="text-purple-600">Privacy Policy</Text>.
        </Text>
```

server/functions/src/index.ts
NasihNazeem: Nah I chose to keep it as `userId` to mirror what we have in `CurrentUser` type. Should I change `userId` in the user object to be `id` as well?

server/functions/src/index.ts
NasihNazeem: The email shouldn't be allowed to be empty since we need an email for the auth table, however, the username can be since that is the next step. This function gets triggered when creating a user for the authentication table and then creates a user in the firestore for that new user.

client/app/(protected)/_layout.tsx
NasihNazeem: Agree, that logic makes sense. Thanks!

server/functions/src/index.ts
JayChuang8: Yeah I think that would be ideal

server/functions/.eslintrc.mjs
NasihNazeem: Yep, I ran `firebase init functions` and it generated a bunch of these files. Do you think I should upgrade it?

client/app/(auth)/signUp.tsx
NasihNazeem: Currently, I set up polling to check when a user verifies their email. This needs to be something that *ideally* checks  when the `emailVerified` value updates when the verify their email, but after taking a closer look, Firebase does not have a listener for email verification just for user sign-in/out. 

Polling is pretty inefficient, so there are some other options. We could consider using a button that checks the verification status and route after that, or we set up a cloud function that we call on the frontend with a listener to check when the emailVerified state gets updated. I do think this is something we should do, but I think we should focus on this down the line. Let me know what you think

client/app/_layout.tsx
NasihNazeem: This is something that React Native Firebase recommends as they have deprecated their API for previous versions (we are on v21). Without this we get a bunch of warning logs

Current ESLint rules don't allow us to move it to the top I think, it needs the imports at the top

https://rnfirebase.io/migrating-to-v22

client/components/AuthListener.tsx
NasihNazeem: Yep, it basically listens to the user state for both sign-in and sign-out.

client/components/AuthListener.tsx
NasihNazeem: Not really this is outdated code. Thanks for noticing, I've updated to the custom spinner

client/app/(auth)/signUp.tsx
NasihNazeem: Yeah it shouldn't be undefined for us. I'll throw in a check and that'll throw an error if it is.

server/functions/.prettierrc
NasihNazeem: this was generated by the CLI command, but I'm gonna replace it to mirror the one in the client folder

client/app/(auth)/signUp.tsx
NasihNazeem: >The SSO login/signup process looks similar between this and the signup screen (eg. UI and callbacks), do you think we can combine it into a component?

I made an SSOButtons component that will handle all SSO buttons and their provider logic. For now we only have one so it seems quite simple, but I think it will be easier to manage when adding Apple down the line. Thanks for the suggestion.

> Should we add something that lets you get to the login screen from this page in case you clicked the wrong thing by accident? For example an "Already have an account?" link. Vice-versa for the login screen

I think it's a good suggestion but I think we should loop Nasrif in before making this change. There is still the possibility of you returning to the onboarding screen to the find your way to the other page. I think we can discuss how best to display this, so I'm not going to implement it now since this can be a quick modification we can add to a later ticket.

client/app/(auth)/forgotPassword.tsx
NasihNazeem: Thanks for noticing this, I added a View component around the conditional error that acts as a spacer similar to what I have in the login screen. It needs to be a separate View because I can't have it under the conditional error as that would only space it appropriately when the error shows, and I also can't have it under the Controller because that would add too much space between the error message and the input field

```suggestion
            <Controller
            control={control}
            name="emailOrUsername"
            render={({ field: { onChange, value } }) => (
              <>
                <Text className="pbk-b2 mb-1.5 text-base-white">
                  Email or username
                </Text>

                <View
                  className={`mb-1 min-h-14 w-full flex-row items-center justify-between rounded-xl border ${errors.emailOrUsername ? "border-red-600" : "border-gray-920"} px-2 py-2`}
                >
                  <TextInput
                    autoCapitalize="none"
                    className="flex-1 text-base-white placeholder:pbk-b1"
                    onChangeText={(text) => {
                      onChange(text);
                    }}
                    placeholder="Enter email or username"
                    placeholderTextColor="gray"
                    value={value}
                  />
                  {errors.emailOrUsername && (
                    <WarningCircle color="red" size={20} weight="bold" />
                  )}
                </View>
              </>
            )}
          />

          <View className="mb-5">
            {errors.emailOrUsername && (
              <Text className="pbk-b3 mb-4 text-red-600">
                {errors.emailOrUsername.message}
              </Text>
            )}
          </View>
```

client/app/(auth)/signUp.tsx
NasihNazeem: I created a `emailVerification.tsx` page that will be the page where users will sit at until they verify their email. If a user abandons the flow directly after sign up, I introduced a check that redirects users to the email verification page until they verify that email. 

If they accidentally provided the wrong email, I also introduced a button to return to the onboarding screen by signing the unverified user out, so that they can start the flow again

client/app/(auth)/signUp.tsx
Dennull: Nice, I like the way it looks now! As discussed offline, we'll address the general routing animation and UI paradigms in a later ticket

client/app/_layout.tsx
Dennull: Oh okay that's fine, we can just leave it there then

client/app/(auth)/login.tsx
Dennull: I tried it out and it does redirect them to the create profile page after logging in. I don't know how this happens tho, do you know what exactly causes the redirect? Is it the `signInWithEmailAndPassword` function?

client/app/(auth)/signUp.tsx
Dennull: Made a new ticket for this so we can track it https://linear.app/rookies-fantasy/issue/ENG-18/establish-error-handling-patterns

client/app/(auth)/signUp.tsx
Dennull: Ohh I see, that makes sense. I agree, polling seems like a janky thing for this so it'd be better if we find a more elegant solution, but we can tackle that later. Made a ticket: https://linear.app/rookies-fantasy/issue/ENG-16/swap-from-polling-for-email-verification

client/app/(auth)/signUp.tsx
Dennull: Nice, this look better on my screen and the markup looks cleaner too 👍

client/app/(auth)/signUp.tsx
Dennull: Nice! For the "already have an account" link, I'm good with waiting on Nasrif's opinion and adding that later, like you said it'd be easy to add in so no rush on that

client/declarations.d.ts
Dennull: Ok, thanks for explaining!

server/functions/.eslintrc.mjs
Dennull: Maybe take a stab at it and it's not a lot of work go for it, but I don't think this should block the PR

client/app/(auth)/emailVerification.tsx
```diff
@@ -0,0 +1,156 @@
...
+          </View>
+
+          <View className="flex-1 justify-end">
+            {cooldown > 0 && (
+              <Text className="pbk-b2 text-center text-base-white">
+                Resend email again in {cooldown} seconds.
+              </Text>
+            )}
+
+            <Pressable
```
Dennull, line 134: Is this supposed to wrap everything? The spacing between the two buttons looks different from the one on the forgot password email sent page too, I think this might also be why

client/app/(auth)/signUp.tsx
```diff
@@ -0,0 +1,234 @@
...
+                SIGN UP
+              </Text>
+            )}
+          </Pressable>
+
+          <Text className="pbk-b1 my-5 flex-row flex-wrap text-gray-600">
+            By signing up, you agree to our{" "}
+            <Text className="text-purple-600">Terms of Service</Text> and{" "}
+            <Text className="text-purple-600">Privacy Policy</Text>.
+          </Text>
```
Dennull: Some minor nitpicks, can we use string interpolation so that we don't need the curly braces (ie. use backticks)? Also, the font size of this seems a little large, should we make it smaller? On Figma it seems like this is font manrope, is that outdated?

client/state/slices/userSlice.ts
```diff
@@ -0,0 +1,35 @@
...
+export const selectCurrentUserId = (state: RootState) => state.user.id;
+
+export const selectIsUserSignedIn = (state: RootState): boolean =>
+  !!state.user.id;
+
+export const selectIsUserRegistered = (state: RootState): boolean =>
+  isNotNil(state.user.id) && isNotNil(state.user.username);
+
+export const selectIsUserVerified = (state: RootState): boolean =>
+  state.user.emailVerified === true;
```
Dennull: Since this is a bool we can just return it as is

server/functions/src/index.ts
```diff
@@ -0,0 +1,25 @@
...
+export const createUserInDatabase = functions.auth
+  .user()
+  .onCreate(async (user: UserRecord) => {
+    const { uid, email, emailVerified } = user;
+    const usersRef = admin.firestore().collection("users");
+
+    try {
+      await usersRef.doc(uid).set({
+        id: uid,
+        email: email,
```
Dennull, line 21: The type of this is `string | undefined`, do you think we should check for undefined that and throw an error if it is? Maybe use a condition like `!!email.trim()`?

If we really wanna get fancy we can cook up some regex to check if it's actually an email as backend validation, but it's not that essential

server/functions/src/index.ts
```diff
@@ -0,0 +1,25 @@
...
+        emailVerified: emailVerified,
+        createdAt: admin.firestore.FieldValue.serverTimestamp(),
+        updatedAt: admin.firestore.FieldValue.serverTimestamp(),
+      });
+      console.log("User created in the users table.");
```
Dennull: Should we drop this?

Dennull (approved):
A few small comments, but I think I'm good to merge after those!

client/app/(auth)/login.tsx
NasihNazeem: The automatic redirects are handled in `auth/_layout.tsx`. Basically, once you sign in using that method Firebase will update the user authentication state. We have a component `AuthListener.tsx` that listens whenever that state is updated, and so if a user is detected we update our local (Redux) user state using `setUser`. 

We use the redux user state in our navigation logic in `auth/_layout.tsx` to determine if a user should be redirected to a certain page based on whether they are registered, verified, etc..

client/app/(auth)/signUp.tsx
NasihNazeem: As discussed offline, this will need to be developed in a future ticket as we do not want to bloat this PR any more

client/app/(auth)/signUp.tsx
NasihNazeem: You are right it's one size too large. I've changed this to `pbk-b2` instead of `pbk-b1`. I also noticed that there was an inconsistency with the password helper text, so I fixed that as well.

Thank you for the string interpolation suggestion as well

client/state/slices/userSlice.ts
NasihNazeem: Since `emailVerified` can be undefined, I'm going to change it to !!state.user.emailVerified

server/functions/src/index.ts
NasihNazeem: I think Firebase handles the regex on their own and validates emails, but I will add a validation check for ourselves if there is something missing.

I'm thinking something like

```
if(!email || !email.trim()) {
    console.error(`${uid} has invalid or missing email`);
    throw new Error("Email is required for user creation.");
}
```

server/functions/.eslintrc.mjs
NasihNazeem: It looks like not all the included plugins support v9 yet

I just took a look at one of them but I think I'll verify with all plugins at a later date to support v9 and once they do we'll convert it

This is a [PR](https://github.com/google/eslint-config-google/pull/77) from google's eslint plugin that firebase generated in the eslint file. Looks like it'll be merged in pretty soon

client/app/(auth)/emailVerification.tsx
NasihNazeem: Definitely not, thanks for catching this. I've also changed the PressableLink here to a regular Pressable as I'm not sure PressableLink is able to handle backwards navigation

client/app/(auth)/login.tsx
Dennull: Ah ok, thanks for explaining!

server/functions/.eslintrc.mjs
Dennull: Ok no worries, we'll upgrade when we can

Dennull (approved):
Made a few small changes, mostly fixing some bugs and slight styling consistency stuff. Take a look at the commits and let me know if you wanna talk about anything. I'm good to merge this!

JayChuang8 (approved):

PR #10, opened 2025-05-19
Feature/ENG-15: User and team setup

JayChuang8:
## Related Issues

Closes ENG-15

## Summary

Add create user profile and create user team screens after login/signup. This includes a date picker field and bottom sheet component.

## Changes Made

- Created new `BottomSheet` component for selecting avatars and team logos
- Created `useController.ts` which contains functions that wrap some common firestore api calls. The functions also map the data retrieved from firestore into our own types that we use for our redux stores
- Added `createTeam` and `createProfile` form screens

## Screenshots

IMG_8468
IMG_8467
IMG_8466
IMG_8465
IMG_8464

## Testing Instructions

No additional notes


client/app/(auth)/createProfile.tsx
```diff
@@ -1,19 +1,430 @@
...
+    ),
+  dateOfBirth: yup.date().required("Date of Birth is required"),
+  username: yup.string().required("Username is required"),
+  name: yup.string().required("Name is required"),
+});
 
 const CreateProfile = () => {
-  const user = useAppSelector((state) => state.user);
+  const auth = getAuth();
+  const currentUser = auth.currentUser;
```
NasihNazeem: I think fetching for the id from the user slice is more fitting since we that is all we use here

```suggestion
  const userId = useAppSelector((state) => state.user.id));
```

client/app/(auth)/createProfile.tsx
```diff
@@ -1,19 +1,430 @@
...
-      <Text className="pbk-h6 text-gray-950">This is temporary</Text>
-      <Text className="pbk-b1 text-gray-950">{user.userId}</Text>
-      <Text className="pbk-b1 text-gray-950">{user.email}</Text>
-      <Text className="pbk-b1 text-gray-950">{user.username}</Text>
-
-      <Link className="mt-4 py-4" href="/">
-        <Text className="pbk-b2 text-purple-500">Go to home screen!</Text>
-      </Link>
+    <View className="flex-1 bg-gray-950">
+      <TouchableWithoutFeedback onPress={Keyboard.dismiss}>
```
NasihNazeem: Dennis mentioned this in my PR but `TouchableWithoutFeedback` seems to be outdated and the docs recommend using `Pressable` moving forward. I noticed that you might still need a `classname="flex-1"` when I was making the suggested change but lmk if you don't

client/app/(auth)/createProfile.tsx
```diff
@@ -1,19 +1,430 @@
...
+              />
+              {errors.dateOfBirth && (
+                <Text className="pbk-b3 mb-4 text-red-600">
+                  {errors.dateOfBirth.message}
+                </Text>
+              )}
+
+              <Text className="pbk-b2 mb-1.5 text-base-white">Avatar</Text>
+              <View className="mb-2 mt-4 flex-row items-center justify-between">
+                <Pressable onPress={() => setShowBottomDrawer(true)}>
```
NasihNazeem: If you have your keyboard open and you open the bottom drawer, the keyboard doesn't seem to dismiss. I think this is because the Pressable captures the touch event and ignores the TouchableWithoutFeedback component. One way to fix this is to explicitly call `Keyboard.dismiss` in the `onPress` of this Pressable. This bug is also there on the "Change avatar" Pressable.

```suggestion
                <Pressable onPress={() => { Keyboard.dismiss(); setShowBottomDrawer(true);}}>
```

client/app/(auth)/createProfile.tsx
```diff
@@ -1,19 +1,430 @@
...
+        <View className="flex-1">
+          <KeyboardAvoidingView behavior="padding" className="flex-1">
+            <View className="flex-1 flex-col px-6 py-4">
+              <View className="my-8 size-8 self-end" />
+
+              <Text className="pbk-h5 mb-8 text-base-white">
+                Tell us who you are
+              </Text>
+
+              <Text className="pbk-b2 mb-1.5 text-base-white">Username</Text>
```
NasihNazeem: I recently made this change in my PR, and it still needs a final review, but what do you think about moving this label into the Controller? I think that makes it a bit more readable as it ties the label to the TextInput. Note: You would need to make this change on all controllers if you agree

```suggestion
              <Text className="pbk-b2 mb-1.5 text-base-white">Username</Text>
              <Controller 
                  ....
                  render= {...} => {
                  <>
                      <Text className="pbk-b2 mb-1.5 text-base-white">Username</Text>
                      <View
                          className={`mb-2 min-h-14 w-full flex-row items-center rounded-xl border             
                          ${errors.username ? "border-red-600" : "border-gray-920"} px-2 py-2`}
                       ></View>
                  </>
```

client/app/(auth)/createProfile.tsx
```diff
@@ -1,19 +1,430 @@
...
+              {errors.username && (
+                <Text className="pbk-b3 mb-4 text-red-600">
+                  {errors.username.message}
+                </Text>
+              )}
+              <View className="mb-5">
+                <Text className="pbk-b3 text-gray-600">
+                  This is your display name. You can change it later.
+                </Text>
+              </View>
```
NasihNazeem: I think we should include the error message into this View you have set up but also create a conditional render for the message so we don't display too many notes. When there is an error that occurs, it should prioritize letting the user know about that error and then go back to the default text. This also fixes some of the spacing issues I think as I changed the margin Bottom of this View to mirror the spacing between the other TextInputs and labels.

```suggestion            
              <View className="mb-4">
                {errors.username ? (
                  <Text className="pbk-b3 text-red-600">
                    {errors.username.message}
                  </Text>
                ) : (
                  <Text className="pbk-b3 text-gray-600">
                    This is your display name. You can change it later.
                  </Text>
                )}
              </View>
```

client/app/(auth)/createProfile.tsx
```diff
@@ -1,19 +1,430 @@
...
+                      </View>
+                    </Pressable>
+
+                    <Modal
+                      animationType="slide"
+                      onRequestClose={() => setShowDatePicker(false)}
+                      transparent
+                      visible={showDatePicker}
+                    >
+                      <TouchableWithoutFeedback
```
NasihNazeem: Just tagging this here but I'd follow the same suggestion as Dennis' mentioned in my PR about changing this to a Pressable with a classname="flex-1"

client/app/_layout.tsx
```diff
@@ -21,7 +21,7 @@ const RootLayoutNav = () => (
         <Provider store={store}>
           <AuthListener>
             <StatusBar />
```
NasihNazeem: Not sure why I don't see it on the other screens, but maybe should consider making this style light

```suggestion
            <StatusBar style="light" />
```

client/app/(auth)/createProfile.tsx
```diff
@@ -1,19 +1,430 @@
...
+    <View className="flex-1 bg-gray-950">
+      <TouchableWithoutFeedback onPress={Keyboard.dismiss}>
+        <View className="flex-1">
+          <KeyboardAvoidingView behavior="padding" className="flex-1">
+            <View className="flex-1 flex-col px-6 py-4">
+              <View className="my-8 size-8 self-end" />
+
+              <Text className="pbk-h5 mb-8 text-base-white">
+                Tell us who you are
+              </Text>
```
NasihNazeem: Instead of having that spacer View component I think we should just add a bit of marginTop to the Text. `mt-24` replicates the spacing you have set up on my screen. Let me know if this works, or if there is another use I am missing for this View component.

```suggestion
              <Text className="pbk-h5 mb-8 mt-24 text-base-white">
                Tell us who you are
              </Text>
```

client/app/(auth)/createProfile.tsx
```diff
@@ -1,19 +1,430 @@
...
+                </Text>
+              </View>
+
+              <Text className="pbk-b2 mb-1.5 text-base-white">Name</Text>
+              <Controller
+                control={control}
+                name="name"
+                render={({ field: { onChange, value } }) => (
+                  <View
+                    className={`mb-4 min-h-14 flex-row items-center justify-between rounded-xl border ${errors.name ? "border-red-600" : "border-gray-920"} px-3 py-2`}
```
NasihNazeem: To keep the spacing consistent, I suggest changing the marginBottom here from `mb-4` to `mb-2` and reflecting this change on the next Controller as well.

client/app/(auth)/createProfile.tsx
```diff
@@ -1,19 +1,430 @@
...
+                      <WarningCircle color="#dc2626" size={20} weight="bold" />
+                    )}
+                  </View>
+                )}
+              />
+              {errors.name && (
+                <Text className="pbk-b3 mb-4 text-red-600">
+                  {errors.name.message}
+                </Text>
+              )}
```
NasihNazeem: Again a spacing suggestion that works when we have an error or not. This is similar to the View you had set up for the username subtext.

```suggestion
             <View className="mb-4">
                {errors.name && (
                  <Text className="pbk-b3 text-red-600">
                    {errors.name.message}
                  </Text>
                )}
              </View>
```

client/app/(auth)/createTeam.tsx
```diff
@@ -0,0 +1,364 @@
...
+    .test(
+      "not-default-logo",
+      "Logo is required",
+      (value) => value !== logoOptions[0].url,
+    ),
+});
+
+const CreateTeam = () => {
+  const auth = getAuth();
+  const currentUser = auth.currentUser;
```
NasihNazeem: Similar to what I mentioned in the createProfile screen, we should fetch for the user id instead of using the getAuth() to find the user since all we really need is the user id which is stored in the slice

client/app/(auth)/createTeam.tsx
```diff
@@ -0,0 +1,364 @@
...
+
+      let teamId: string;
+
+      if (teams?.length === 0) {
+        teamId = await UserController.addUserTeam(userId, model);
+      } else if (teams?.length === 1) {
+        teamId = teams[0].id;
+        await UserController.editUserTeam(userId, teamId, model);
+      } else {
+        throw new Error("Unexpected number of teams in create team screen");
```
NasihNazeem: nice error being thrown here, good stuff

client/app/(auth)/createTeam.tsx
```diff
@@ -0,0 +1,364 @@
...
+    } catch (error) {
+      console.error("Failed to create team:", error);
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  return (
+    <View className="flex-1 bg-gray-950">
+      <TouchableWithoutFeedback onPress={Keyboard.dismiss}>
```
NasihNazeem: tagging this to follow the suggestion mentioned in the `createProfile` page

client/app/(auth)/createTeam.tsx
```diff
@@ -0,0 +1,364 @@
...
+                <Pressable
+                  className="mr-3"
+                  onPress={() => router.push("/(auth)/createProfile")}
+                >
+                  <BackIcon fill="white" height={25} width={25} />
+                </Pressable>
+                <Text className="pbk-h5 mb-8 text-base-white">
+                  Create your team
+                </Text>
+              </View>
```
NasihNazeem: To remove the spacer and fix up the icon and title spacing, and replace the arrow svg with a icon from phosphor:
```suggestion
              <View className="mb-8 mt-24 flex-row items-center gap-4">
                <Pressable
                  className="size-8 items-center justify-center rounded-md border border-gray-900 p-4"
                  onPress={() => router.push("/(auth)/createProfile")}
                >
                  <ArrowLeft color="white" size={20} weight="bold" />
                </Pressable>
                <Text className="pbk-h5 text-base-white">Create your team</Text>
              </View>
```

client/app/(auth)/createTeam.tsx
```diff
@@ -0,0 +1,364 @@
...
+              />
+              {errors.abbreviation && (
+                <Text className="pbk-b3 mb-4 text-red-600">
+                  {errors.abbreviation.message}
+                </Text>
+              )}
+
+              <Text className="pbk-b2 mb-1.5 text-base-white">Team logo</Text>
+              <View className="mb-2 mt-4 flex-row items-center justify-between">
+                <Pressable onPress={() => setShowBottomDrawer(true)}>
```
NasihNazeem: I mentioned a bug in the `createProfile` page related to the keyboard and provided a solution that could also fix the issue here. But let me know if you like the explicit call or not

client/app/(auth)/login.tsx
```diff
@@ -102,14 +95,24 @@ const Login = () => {
...
+
+      if (!isUserRegistered) {
+        router.replace("/(auth)/createProfile");
+      }
+
+      if (!isTeamRegistered) {
+        router.replace("/(auth)/createTeam");
+      }
+
+      router.replace("/(protected)");
```
NasihNazeem: We don't need these `router.replace()` calls as this should be handled at the _layout.tsx level

client/app/(auth)/login.tsx
NasihNazeem, line 1: Great job making that controllers folder to handle all the querying. You made a few modifications here related to that and the form types.

client/app/(auth)/signUp.tsx
NasihNazeem, line 1: You won't find it here but once you pull the most recent changes you will see a `emailVerification` page that has a query on there updating the emailVerified field of a user. Please handle that at the controller level as well, thank you

I could also do it on my branch if we end up merging this branch first

client/assets/icons/back-icon.svg
NasihNazeem, line 1: We can remove this

client/assets/images/back-icon.png
NasihNazeem, line 1: We can remove this

client/components/AuthListener.tsx
NasihNazeem, line 1: This change is amazing! Thanks for this

client/controllers/userController.ts
NasihNazeem, line 1: Looks good to me

client/state/slices/teamSlice.ts
```diff
@@ -0,0 +1,24 @@
...
+  },
+});
+
+export const selectTeamId = (state: RootState) => state.team.id;
+
+export const selectIsTeamRegistered = (state: RootState): boolean =>
+  state.team.id !== undefined &&
+  state.team.abbreviation !== undefined &&
+  state.team.logoUrl !== undefined &&
+  state.team.name !== undefined;
```
NasihNazeem: We could use that `isNotNil` util func here

Do we need to check if all things are undefined or can we just check for one of these (team.id) optimistically? I assume the others should not exist if we don't have an id

client/state/slices/userSlice.ts
```diff
@@ -1,31 +1,27 @@
...
-  typeof state.user.userId === "string" && state.user.userId.length > 0;
+  typeof state.user.id === "string" && state.user.id.length > 0;
 
-export const selectIsRegistered = (state: RootState): boolean =>
-  state.user.userId !== undefined && state.user.username !== undefined;
+export const selectIsUserRegistered = (state: RootState): boolean =>
+  state.user.id !== undefined &&
+  state.user.username !== undefined &&
+  state.user.dateOfBirth !== undefined &&
+  state.user.avatarUrl !== undefined;
```
NasihNazeem: similar thought process for me here as mentioned in the teamSlice.ts feeback

NasihNazeem (changes requested):
This looks amazing! I left a few suggestions with many of those suggestions up for discussion. 

I'm making this review after pushing up a bunch of changes on my branch so you might need to look at my stuff before fixing any conflicts, just to make sure you're not pulling anything that will be changed again

client/app/(auth)/createProfile.tsx
Dennull: Honestly this would be nice to get in a common form component so we don't have to keep remembering. I made at ticket for form standardization: https://linear.app/rookies-fantasy/issue/ENG-14/standardize-forms

My only actual comment on this specifically is to double check that the spacing is consistent between the fields whatever we decide to do

client/app/(auth)/_layout.tsx
Dennull, line 1: Technically the registration process happens after auth, so should we move `createProfile.tsx`, `createTeam.tsx`, and the associated routing logic into `(protected)`?

client/app/(auth)/createProfile.tsx
```diff
@@ -1,19 +1,430 @@
...
+import { setUser } from "@/state/slices/userSlice";
+
+type AvatarOption = {
+  url: string;
+  source: ImageSourcePropType;
+};
+
+const avatarOptions: AvatarOption[] = [
+  {
+    url: "../../assets/images/placeholder-avatar.png",
```
Dennull: Should we swap to using ids to make things simpler to read?

client/app/(auth)/createProfile.tsx
```diff
@@ -1,19 +1,430 @@
...
+    const setDefaultUserData = async () => {
+      if (currentUser) {
+        const userData = await UserController.getUser(currentUser.uid);
+        if (userData) {
+          const matchedAvatar = avatarOptions.find(
+            (option) => option.url === userData!.avatarUrl,
+          );
+          if (matchedAvatar) {
+            setSelectedAvatarOption(matchedAvatar);
+          }
```
Dennull: Nit, do we need the `!` operator? Also, do you think dropping the second if statement and defaulting to the first avatar option if they didn't select one yet works?

```suggestion
          const matchedAvatar = avatarOptions.find(
            (option) => option.url === userData.avatarUrl,
          );
          setSelectedAvatarOption(matchedAvatar || avatarOptions[0]);
```

client/app/(auth)/createProfile.tsx
```diff
@@ -1,19 +1,430 @@
...
+  const handleCreateProfile = async (model: UserEditModel) => {
+    setIsLoading(true);
+    try {
+      if (currentUser) {
+        await UserController.editUser(currentUser.uid, model);
+        const userData = await UserController.getUser(currentUser.uid);
+        dispatch(setUser(userData));
+        router.push("/(auth)/createTeam");
+      } else {
+        router.push("/(auth)");
```
Dennull: Do we need this else?

client/app/(auth)/createProfile.tsx
```diff
@@ -1,19 +1,430 @@
...
-    <View className="flex-1 items-center justify-center p-5">
-      <Text className="pbk-h6 text-gray-950">This is temporary</Text>
-      <Text className="pbk-b1 text-gray-950">{user.userId}</Text>
-      <Text className="pbk-b1 text-gray-950">{user.email}</Text>
-      <Text className="pbk-b1 text-gray-950">{user.username}</Text>
-
-      <Link className="mt-4 py-4" href="/">
-        <Text className="pbk-b2 text-purple-500">Go to home screen!</Text>
-      </Link>
+    <View className="flex-1 bg-gray-950">
```
Dennull: The wrapper doesn't have scroll enabled, so on small screens if there are form errors or if things overflow you can't scroll to the bottom to hit the "continue" button. Maybe we need to use `ScrollView`?

client/app/(auth)/createProfile.tsx
Dennull: I agree, I don't like the extra space on the top. I would rather the content of the form start at the top with a bit of padding between the top of the screen, and then let it be as tall as it needs. Any extra white space should be pushed to in between the form and the continue button, similar to how it is in Nasrif's design

client/app/(auth)/createProfile.tsx
```diff
@@ -1,19 +1,430 @@
...
+                  {errors.username.message}
+                </Text>
+              )}
+              <View className="mb-5">
+                <Text className="pbk-b3 text-gray-600">
+                  This is your display name. You can change it later.
+                </Text>
+              </View>
+
+              <Text className="pbk-b2 mb-1.5 text-base-white">Name</Text>
```
Dennull: Are we dropping this form field still? We talked about not needing the full name, but I can't remember if we decided on something

client/app/(auth)/createProfile.tsx
```diff
@@ -1,19 +1,430 @@
...
+              <View className="mb-2 mt-4 flex-row items-center justify-between">
+                <Pressable onPress={() => setShowBottomDrawer(true)}>
+                  <Image
+                    className="h-24 w-24 rounded-full border-2 border-purple-600"
+                    source={selectedAvatarOption.source}
+                  />
+                </Pressable>
+                <Pressable onPress={() => setShowBottomDrawer(true)}>
+                  <Text className="text-purple-600">Change avatar</Text>
+                </Pressable>
```
Dennull: Two things here, this button seems a bit small, it's hard to press sometimes. Do you think we should add padding or bump up the text size a bit? This font doesn't seem styled either

Other thing is this text is sentence case while the one on the create team one is title case (ie. "Create Team Logo"), we should choose one and stick to it. I personally like sentence more but whichever works

client/app/(auth)/createTeam.tsx
```diff
@@ -0,0 +1,364 @@
...
+    setDefaultTeamData();
+  }, [currentUser, reset]);
+
+  const handleCreateTeam = async (model: TeamEditModel) => {
+    setIsLoading(true);
+    try {
+      if (!currentUser) {
+        router.push("/(auth)");
+        return;
+      }
```
Dennull: Do we need this?

client/app/(auth)/createTeam.tsx
Dennull, line 1: We made a lot comments on the create profile page, and I didn't wanna repeat a bunch of them haha so just a reminder that those may apply here as well since the pages are so similar

client/components/BottomSheet.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+  const allSnapPoints = [
+    ...new Set(["0%", ...(snapPoints ?? defaultSnapPoints)]),
+  ].map((o) => snapPointMap[o as SnapPoint]);
+
+  useEffect(() => {
+    if (isOpen) {
+      backdropOpacity.value = withTiming(1, { duration: 200 });
+      setVisible(true);
+      scrollTo(allSnapPoints[1] ?? snapPointMap["33%"]);
```
Dennull, line 93: Is `allSnapPoints` ordered? This probably doesn't matter since the devs pass the `snapPoints` prop in, but what if someone adds snap points out of order, does it break?

client/components/BottomSheet.tsx
Dennull, line 1: This looks great!!

client/app/(auth)/createTeam.tsx
```diff
@@ -0,0 +1,364 @@
...
+                </Text>
+              )}
+              <View className="mb-5">
+                <Text className="pbk-b3 text-gray-600">
+                  This is your team name. You can change it later.
+                </Text>
+              </View>
+
+              <Text className="pbk-b2 mb-1.5 text-base-white">
+                Team Abbreviation
```
Dennull: Should we add a helper message saying that it has to be 3 letters?

Dennull (changes requested):

client/state/slices/teamSlice.ts
JayChuang8: We didn't have any validation/checks in firestore to ensure they're all filled in which is why I had all these checks to enforce it. But yeah it's possible to just check for the team id, I added a rule in Firestore so that teams can only be written (create, update) if all of these fields are filled in.

client/components/BottomSheet.tsx
JayChuang8: Nah it's not ordered, whichever snappoint is passed in first in the array will be the snappoint used when the sheet is opened. It won't break if it's out of order, as it's just a list of all the possible positions that the bottomsheet can be moved to.

client/app/(auth)/createTeam.tsx
JayChuang8: Cool, this worked great, thanks!

client/app/_layout.tsx
JayChuang8: Changed!

client/app/(auth)/login.tsx
JayChuang8: Ok cool thanks, removed

client/app/(auth)/createProfile.tsx
JayChuang8: Nah, removed now

client/app/(auth)/createProfile.tsx
JayChuang8: No we don't need it since there is the if statement wrapping it, good catch! Removed and changed to the suggested default logic

client/app/(auth)/createProfile.tsx
JayChuang8: I think this should be a problem for another day, when we make common screen layouts with headers and footers. On top of these complexities with adding the `ScrollView`, we would have to apply it to other screens as well, so I think we should exclude it from this PR.

client/app/(auth)/createProfile.tsx
JayChuang8: Yeah makes sense, changed to fetch from the user slice now

client/app/(auth)/createProfile.tsx
JayChuang8: Changed to use `Pressable` and flex!

client/app/(auth)/createProfile.tsx
JayChuang8: Ok, I changed it now so the `View` spacer is removed and the margin is moved to the `Text` component. I reduced it a bit to `16px` so move it closer to the top of the screen.

client/app/(auth)/createProfile.tsx
JayChuang8: Yeah I think that's a good idea. I changed it for all the other for Controllers as well now. I think in the future we can eventually just wrap the `Controller` component and pass in a `label` prop, where it is automatically set in the `Text` component

client/app/(auth)/createProfile.tsx
JayChuang8: Yeah nice, this works well! Changed now

client/app/(auth)/createProfile.tsx
JayChuang8: Yeah I just dropped it, we won't have any use for the user's full name I don't think. There's also no validation for making sure it's actually a full name so people would be able to just put anything in

client/app/(auth)/createProfile.tsx
JayChuang8: Nice catch, your solution works great!

client/app/(auth)/createProfile.tsx
JayChuang8: I added some padding to it and changed the font to match with the form labels. I don't want to increase the font size or it will look a bit off compared to the form labels. Also I'll change the team on to sentence case as well.

client/app/(auth)/createProfile.tsx
JayChuang8: Changed!

client/app/(auth)/createTeam.tsx
JayChuang8: No not anymore, not that we are getting the userId from the user slice. Removed now

client/app/(auth)/createTeam.tsx
JayChuang8: I added a `maxLength={3}` on the text input now, so I don't think we need a message

client/app/(auth)/createTeam.tsx
JayChuang8: Yep, fixed here as well now

client/app/(auth)/createProfile.tsx
JayChuang8: Like rename the png file names to just the id? I've changed it now to that, let me know if you meant something else

client/app/(auth)/signUp.tsx
JayChuang8: Changed in `emailVerification.tsx` to use the controller now!

client/app/(auth)/_layout.tsx
NasihNazeem: As discussed offline, we will be moving the `createProfile.tsx`, `createTeam.tsx` and the associated routing logic into `(protected)`. @JayChuang8 You might have to edit all of the expo router navigation calls associated with those pages.

NasihNazeem (approved):

client/app/(auth)/createProfile.tsx
Dennull: Made a new ticket for this: https://linear.app/rookies-fantasy/issue/ENG-28/explore-making-common-layout-components-and-using-scrollview

client/app/(auth)/createProfile.tsx
Dennull: Oh nah I meant like instead of using the `url` as the identifier in the `.find()` we change it to be an id. Now that this has been merged I can maybe do it in my PR

PR #11, opened 2025-06-16
Feature/ENG-20: Create and populate static data (NBA teams, players, and 2024/25 player averages)

NasihNazeem:
## Related Issues

Closes ENG-20.

## Summary

This PR introduces a few scripts that helped populate the firestore with the data required to help build up the next few features for release. It also introduces a new folder called `scripts` at the root level to keep any scripts created separate to both the server and client as it felt like something that didn't belong in either folder.

## Changes Made

- Created a new folder `scripts` at the root level
- Populated `nbaTeams`, `nbaPlayers` and `nbaPlayerAverages` collections using the BallDontLie SDK

## Screenshots

No additional screenshots

## Testing Instructions

No additional notes


scripts/fillNBAPlayerAverages.js
```diff
@@ -0,0 +1,93 @@
...
+
+const chunkArray = (arr, size) => {
+  const result = [];
+  for (let i = 0; i < arr.length; i += size) {
+    result.push(arr.slice(i, i + size));
+  }
+  return result;
+};
+
+const normalizeAverages = (raw) => ({
```
Dennull, line 17: Nit, is this really "normalizing"? It's more like mapping the raw data, isn't it? Should we rename it to `mapAverages` or something like that? Or maybe there's a meaning to normalize that I'm not aware of

scripts/fillNBAPlayers.js
```diff
@@ -0,0 +1,74 @@
...
+      totalStored += players.length;
+
+      console.log(
+        `Stored ${players.length} players. Total so far: ${totalStored}`
+      );
+      cursor = meta?.next_cursor;
+    } while (cursor);
+
+    await batch.commit();
+    console.log(`Finished storing NBA players. Total stored: ${totalStored}`);
```
Dennull, line 68: How come these don't get the check mark and x emojis 😂

scripts/fillNBATeams.js
```diff
@@ -0,0 +1,52 @@
...
+      city: team.city,
+      conference: team.conference,
+      division: team.division,
+      fullName: team.full_name,
+      name: team.name,
+      state: "",
+      logoUrl: "",
+    }));
+
+    console.log(normalizedTeams);
```
Dennull: Nit, do we need this console log?

scripts/fillNBATeams.js
```diff
@@ -0,0 +1,52 @@
...
+    if (!teams) {
+      console.error("No teams to store.");
+      return;
+    }
+
+    const filteredTeams = teams.filter(
+      (team) => team.division && team.division.trim() !== ""
+    );
+
+    const normalizedTeams = filteredTeams.map((team) => ({
```
Dennull: Another nit, but do you think we should make this a function like in the other scripts and use it when we do the `foreach` on line 39?

Alternatively if you wanted, you could also extract this into a function and pass it directly into the map like `filteredTeams.map(normalizeTeams)`

scripts/migrateAverages.js
Dennull, line 1: Wait I'm a bit lost, can you please give me a brief explanation of what this script does?

NasihNazeem: No worries, to start — loading player data with player averages firebase requires me to create fetches for each (`nbaPlayers` and `nbaPlayerAverages`) collection. Firebase doesn't support anything similar to a "join" at the collection level, so I needed to find a way to get the data in one query. The recommended way actually was to duplicate the data I needed into one collection. I was fine with it in this case since nba player averages are related 1-to-1 to an nba player, so this script helped me migrate all the fields from the `nbaPlayerAverages` collection into the `nbaPlayers` collection.

I decided to keep the `nbaPlayerAverages` collection though because this could contain historical data as we also hold a `seasonId` in there.

This reduced the amount of reads to the firestore by a lot

scripts/fillNBATeams.js
NasihNazeem: I initially wanted to keep it but I think overall we have another console.log proving if the script was successful so I think I'll just remove this

scripts/fillNBAPlayers.js
NasihNazeem: lmao GPT didn't have me covered on this one

I'm going to remove them from the other

scripts/fillNBAPlayerAverages.js
NasihNazeem: I'm fine with both, but I guess I called it normalizing because I was reshaping the data, but I think this func could also be called `mapAverages`. I don't have a preference so I'll make the change!

scripts/fillNBATeams.js
NasihNazeem: yeah thats a good idea, updated

JayChuang8 (approved):

PR #12, opened 2025-07-20
Feature/ENG-29, ENG-31: Remove gluestack and fix user creation routing issues

Dennull:
## Related Issues

Addresses ENG-29 and ENG-31

## Summary

This PR removes gluestack from our codebase (as it was unused), fixed a bug with routing during user creation, and introduces better handling of safe areas

## Changes Made

I noticed a few tech debt things while working on another ticket, so I decided to open a PR to fix them while I was at it:

- I removed gluestack-ui from our app since it's not really used. We liked the control of making our own components and meant to remove it but have been putting it off, so I figured I'd do it now
- A bug was fixed for when you tried to press the back button on the "Create Team" page to return to the "Create Profile" page after saving your profile details and refreshing the app
  - The back button wouldn't work because when you refreshed the page, you were immediately taken to the create profile page, but the history stack got cleared, `router.back()` wouldn't work
  - I tried to get something working as it was originally intended, but it ended up being a time sink so instead I opted to always send the user to the create profile page first
- Our routing logic was updated to use guards only and remove some existing uses of the `Redirect` component; if we can get away with using one pattern I think that's better
- Finally, our existing pages were updated to handle safe areas better. This mainly involved wrapping the entire page with the `SafeAreaView` component from `react-safe-area-context`

## Screenshots

No additional screenshots

## Testing Instructions

No additional notes


NasihNazeem (approved):
Tested the changes, looks good to me. There wasn't much to review as you ensured all pages have `SafeAreaView` and removed all gluestack related files and implementations.

Thanks!

JayChuang8 (approved):

PR #13, opened 2025-07-22
Feature/ENG-7: Create team builder player table UI and data fetching

NasihNazeem:
## Related Issues

Closes ENG-7.

## Summary

This PR introduces a player table for the team building portion of the app. This is where players get selected to create a user's initial lineup. This table showcases all players ordered by salary as well as their player data including statistics. Users are not able add or remove players to the lineup in this PR but that logic will be addressed in a following ticket. 

## Changes Made

- Introduced an IconButton and changed the sizing of other buttons to match that sizing
- Created draft folder inside protected screens to handle different tab layout for team building
- Created custom table component to handle sticky and scrollable columns
- Implemented infinite scroll within players table component using Tanstack Query
- Team balance is working programmatically (have not implemented logic yet)
- Created a controller to handle fetches for nba players

One thing I did not include in this PR is the tier badge for each player. I think I want to include this in a separate ticket as we have no use for it until those filters come in.

## Screenshots

https://github.com/user-attachments/assets/6af2ea63-dd26-46d6-a629-2295de076b97

## Testing Instructions

Each fetch for the nba players takes a few hundred reads so just keep that in mind when testing! :)


client/app/(auth)/forgotPassword.tsx
```diff
@@ -97,7 +97,7 @@ const ForgotPassword = () => {
         >
           <View className="my-20 flex-row items-center gap-4">
             <Pressable
-              className="size-8 items-center justify-center rounded-md border border-gray-900 p-4"
+              className="size-10 items-center justify-center rounded-md border border-gray-900 p-4"
               onPress={() => router.dismissAll()}
             >
               <ArrowLeft color="white" size={20} weight="bold" />
```
JayChuang8: Should we use the new `IconButton` here as well?

client/app/(auth)/login.tsx
```diff
@@ -101,7 +101,7 @@ const Login = () => {
         >
           <View className="my-20 flex-row items-center gap-4">
             <Pressable
-              className="size-8 items-center justify-center rounded-md border border-gray-900 p-4"
+              className="size-10 items-center justify-center rounded-md border border-gray-900 p-4"
               onPress={() => router.dismissAll()}
             >
               <ArrowLeft color="white" size={20} weight="bold" />
```
JayChuang8: Should we use the new `IconButton` here as well?

client/app/(auth)/signUp.tsx
```diff
@@ -103,7 +103,7 @@ const SignUp = () => {
         <KeyboardAvoidingView behavior="padding" className="flex-col px-6 py-4">
           <View className="my-20 flex-row items-center gap-4">
             <Pressable
-              className="size-8 items-center justify-center self-start rounded-md border border-gray-900 p-4"
+              className="size-10 items-center justify-center self-start rounded-md border border-gray-900 p-4"
               onPress={() => router.back()}
             >
               <ArrowLeft color="white" size={20} weight="bold" />
```
JayChuang8: Should we use the new `IconButton` here as well?

client/controllers/nbaPlayersController.ts
```diff
@@ -0,0 +1,69 @@
+import firestore, {
+  FirebaseFirestoreTypes,
+} from "@react-native-firebase/firestore";
+import { Player } from "@/types/players";
+
+const PLAYERS_COLLECTION = "nbaPlayers";
+
+export class NBAPlayersController {
```
JayChuang8: Minor nit, can we use pascal case here? `NbaPlayersController`

client/app/(protected)/(draft)/players.tsx
```diff
@@ -0,0 +1,220 @@
...
+};
+
+const Players = () => {
+  const [query, setQuery] = useState("");
+  const router = useRouter();
+
+  const fetchPlayersWithAverages = async ({
+    pageParam,
+  }: FetchPlayersParams = {}) => {
+    const PAGE_SIZE = 25;
```
JayChuang8: Can we put this const outside the component? It's minor but this way it doesn't get recreated on every render. Then this function will be simplified too, we can get rid of the return and just do `() => await NBAPlayersController.getFreeAgents(PAGE_SIZE, pageParam)`

client/app/(protected)/createTeam.tsx
```diff
@@ -180,7 +181,7 @@ const CreateTeam = () => {
             <View className="flex-1 flex-col px-6 py-4">
               <View className="mb-8 mt-20 flex-row items-center gap-4">
                 <Pressable
-                  className="size-8 items-center justify-center rounded-md border border-gray-900"
+                  className="size-10 items-center justify-center rounded-md border border-gray-900"
                   onPress={() => router.back()}
                 >
                   <ArrowLeft color="white" size={20} weight="bold" />
```
JayChuang8, line 187: Should we use the new `IconButton` here as well?

client/controllers/nbaPlayersController.ts
```diff
@@ -0,0 +1,69 @@
...
+  static getFreeAgents = async (
+    PAGE_SIZE: number,
+    pageParam?: FirebaseFirestoreTypes.DocumentSnapshot,
+  ): Promise<{
+    players: Player[];
+    lastDoc?: FirebaseFirestoreTypes.DocumentSnapshot;
+    hasMore: boolean;
+  }> => {
```
JayChuang8: Minor nit, can we make a type for this return payload please?

client/controllers/userController.ts
```diff
@@ -113,6 +115,8 @@ export class UserController {
         id: team.id,
         logoUrl: team.data()?.logoUrl,
         name: team.data()?.name,
+        lineup: team.data()?.lineup,
```
JayChuang8: I noticed there is lineup here but not anywhere else.  Do we want it in this PR or leave it for a future one?

JayChuang8 (changes requested):
Looks very nice, great work!

client/controllers/userController.ts
NasihNazeem: good catch, I was working on the lineup logic initially on this ticket but decided to remove everything. Looks like I missed this. I will remove this

client/app/(protected)/createTeam.tsx
NasihNazeem: yes good idea

client/app/(protected)/(draft)/players.tsx
NasihNazeem: yeah totally, good catch. I will make that adjustment

client/controllers/nbaPlayersController.ts
NasihNazeem: yes thank you

client/controllers/nbaPlayersController.ts
NasihNazeem: Yes, I'm going to make a return type named `PlayerFetchResult` and use that. Thank you

client/app/(protected)/(draft)/players.tsx
```diff
@@ -0,0 +1,223 @@
...
+                "REB",
+                "AST",
+                "STL",
+                "BLK",
+                "TO",
+                "FPTS",
+              ]}
+              isFetchingNextPage={isFetchingNextPage}
+              onEndReached={() => fetchNextPage()}
+              stickyIndexes={[0, 1]}
```
JayChuang8: I think we should just pass in a number n, for the first n columns to be sticky.  In `Table.tsx` we're slicing everything from 0 anyways so only the first 0-n columns can be sticky. This way we don't have to do `sitckyIndexes.length` when slicing

client/components/Table/Row.tsx
```diff
@@ -0,0 +1,32 @@
+import React from "react";
+import { View } from "react-native";
+import Cell from "./Cell";
+
+type RowProps = {
+  rowData: any[];
+  stickyIndexes?: number[];
```
JayChuang8: This prop is unused in this component, I don't think we need it

client/components/Table/Table.tsx
```diff
@@ -0,0 +1,140 @@
...
+    isSyncingRef.current = true;
+
+    let frameId: number | null = null;
+    if (source === "header") {
+      bodyScrollRef.current?.scrollTo({ x, animated: false });
+    } else {
+      horizontalScrollRef.current?.scrollTo({ x, animated: false });
+    }
+
+    if (frameId !== null) cancelAnimationFrame(frameId);
```
JayChuang8: It looks like this is dead code, it never runs because frameId is always set to null on line 42

client/components/Table/Table.tsx
```diff
@@ -0,0 +1,140 @@
...
+          rowData={stickyHeaders}
+          variant="header"
+          widthClasses={stickyWidths}
+        />
+        <ScrollView
+          bounces={false}
+          horizontal
+          onScroll={(e) => {
+            const x = e.nativeEvent.contentOffset.x;
+            handleHorizontalScroll(x, "header");
```
JayChuang8: Minor nit, can we just do `handleHorizontalScroll(e.nativeEvent.contentOffset.x, "header")`?

client/types/teamTypes.ts
```diff
@@ -3,6 +3,10 @@
   id: string;
   logoUrl?: string;
   name?: string;
+  balance: number;
 };
 
-export const defaultTeam: Team = { id: "" };
+export const defaultTeam: Team = {
+  id: "",
+  balance: 100000000,
```
JayChuang8: I think we should have this as a const in this file since it's reused in `userController.ts`

NasihNazeem: Sounds good. I'll name it `TEAM_BALANCE` and also be updating it to 150M as suggested by Nasrif

client/controllers/nbaPlayersController.ts
```diff
@@ -0,0 +1,71 @@
...
+const PLAYERS_COLLECTION = "nbaPlayers";
+
+type PlayerFetchResult = {
+  players: Player[];
+  lastDoc?: FirebaseFirestoreTypes.DocumentSnapshot;
+  hasMore: boolean;
+};
+
+export class NbaPlayersController {
+  static getFreeAgents = async (
```
JayChuang8: Minor nit, can we rename this to getPlayers? Eventually when we want to only get free agents we can just filter from this list

client/controllers/nbaPlayersController.ts
```diff
@@ -0,0 +1,71 @@
...
+    PAGE_SIZE: number,
+    pageParam?: FirebaseFirestoreTypes.DocumentSnapshot,
+  ): Promise<PlayerFetchResult> => {
+    try {
+      let query = await firestore()
```
JayChuang8: The `await` can be removed here since it's not actually executing, it's just building the query

client/components/Table/Row.tsx
NasihNazeem: Good catch

client/controllers/nbaPlayersController.ts
NasihNazeem: oh yeah good catch. I moved the get() underneath it.

client/controllers/nbaPlayersController.ts
NasihNazeem: Sounds good

client/components/Table/Table.tsx
NasihNazeem: good catch Jay. Thanks

client/app/(protected)/(draft)/players.tsx
NasihNazeem: Okay that makes sense. I've renamed it to be `stickyColumns` that takes in n columns. I followed your advice on applying it within the table component as well.

Thanks for the suggestion, this solves one of my own gripes with this component lol

client/components/Table/Table.tsx
NasihNazeem: I ended up removign the frameId variable and the animation frame functions as I don't think it's necessary.

client/controllers/userController.ts
```diff
@@ -66,6 +66,7 @@ export class UserController {
         .collection(TEAMS_COLLECTION)
         .add({
           ...params,
+          balance: 100000000,
```
JayChuang8: Let's use `TEAM_BALANCE` here too

NasihNazeem: Good catch, thank you

JayChuang8 (approved):

PR #14, opened 2025-07-25
Feature/ENG-26: Team augments

JayChuang8:
## Related Issues

Closes ENG-26

## Summary

Added team augments page, listing 4 random augments. The random shuffle will be replaced eventually when we figure out where we want to put the shuffle logic. We will also store the user's chosen augment in the team collection later as well once we work out how we want to clear it every week

## Changes Made

- Seed the database with augment URLs and descriptions
  - We should look into blob storage soon for our assets, either S3 or probably the Firebase Cloud Storage since we are already deep into using Google's stuff: [https://firebase.google.com/docs/storage](url)
- Added the `react-native-linear-gradient` library for gradient borders around augment cards
- Refactored buttons; abstracted common styling and logic into a `Button.tsx` component so the styles are consistent across all buttons
- Added a `Dialog.tsx` wrapper component around react native modal

## Screenshots

IMG_29C48E313CBB-1

IMG_9327
IMG_9328

## Testing Instructions

No additional notes


client/assets/images/augments/background-image.png
Dennull, line 1: Just more of a general comment, it's not a huge deal now cause these pictures are pretty small but eventually we'd want to prevent having so many images to reduce our bundle size right? I'm not sure how other apps do it, do they store it online and fetch for it when they need?

client/components/Button.tsx
```diff
@@ -0,0 +1,43 @@
+import { Pressable, Text } from "react-native";
+import Spinner from "./Spinner";
+
+type ButtonProps = {
+  className?: string;
+  disabled?: boolean;
+  isLoading?: boolean;
+  onPress?: () => void;
+  text?: string;
```
Dennull: Do you think "label" sounds better? Also, should this be required? Doesn't really make sense to have a button with no text

client/components/Dialog.tsx
```diff
@@ -0,0 +1,41 @@
+import { X } from "phosphor-react-native";
+import React from "react";
+import { Modal, Pressable, Text, View } from "react-native";
+import Button from "./Button";
+
+type DialogProps = {
+  children: React.ReactNode;
+
```
Dennull: Nit, extra whitespace here

client/app/(protected)/(draft)/applyAugment.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+  useEffect(() => {
+    const fetchAugments = async () => {
+      try {
+        setIsLoading(true);
+        const data = await AugmentController.getAugments();
+
+        // Shuffle the array and take the first 4
+        // TODO: Replace this random selection logic with a more meaningful selection process that persists weekly in the backend.
+        const shuffled = data.sort(() => 0.5 - Math.random());
```
Dennull, line 81: Yeah I asked chatgpt about this and it said that what we have can be an unreliable shuffle because of the JavaScript algorithm. It's probably not the best if we want to keep things completely fair because some outcomes may be more likely than others

In the long term we should probably try to find a well established shuffling algorithm to use. Chatgpt suggested the Fisher-Yates (aka Knuth) algorithm, but we should do a bit of our own research too. This should not be a blocker for this PR but can you also update the TODO above to explicitly mention that the shuffle algorithm is what we should look into?

client/app/(protected)/(draft)/applyAugment.tsx
```diff
@@ -0,0 +1,229 @@
...
+    } catch (error) {
+      console.error("Failed to set augmentId:", error);
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  return (
+    <View className="flex-1 bg-gray-950">
+      <Pressable className="flex-1" onPress={Keyboard.dismiss}>
```
Dennull: We should probably make a component for this at some point lol

client/app/(protected)/(draft)/applyAugment.tsx
```diff
@@ -0,0 +1,229 @@
...
+      router.replace("/(protected)/(tabs)");
+    } catch (error) {
+      console.error("Failed to set augmentId:", error);
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  return (
+    <View className="flex-1 bg-gray-950">
```
Dennull: In my recent PR I started using `SafeAreaView` from `react-native-safe-area-context`, I think we should start using that for most pages to handle safe areas well

client/app/(protected)/(draft)/applyAugment.tsx
```diff
@@ -0,0 +1,229 @@
...
+            <View className="flex-1 flex-col px-6 py-4">
+              <View className="mb-4 mt-20 flex-row items-center gap-4">
+                <Pressable
+                  className="size-8 items-center justify-center rounded-md border border-gray-900"
+                  onPress={() => router.back()}
+                >
+                  <ArrowLeft color="white" size={20} weight="bold" />
+                </Pressable>
+                <Text className="pbk-h5 text-base-white">
+                  Select one augment
```
Dennull: I feel like "an" sounds better than "one", do you agree?

client/app/(protected)/(draft)/applyAugment.tsx
```diff
@@ -0,0 +1,229 @@
...
+                                <Text className="pbk-b3 text-center text-base-white">
+                                  {card.description}
+                                </Text>
+                                <Text className="pbk-b3 mt-2 text-center text-green-400">
+                                  {card.info}
+                                </Text>
+                              </View>
+
+                              <View className="flex-row justify-end">
+                                <CheckCircle
```
Dennull: In general do you think we should style icons using the phosphor props, or className to be consistent with everything else? I don't think it matters that much but @NasihNazeem @JayChuang8 do either of you have any preferences?

client/app/(protected)/(tabs)/index.tsx
```diff
@@ -25,12 +26,12 @@ const Home = () => {
...
       <View className="my-30 h-16" />
-      <Pressable
-        className="min-h-12 w-full justify-center rounded-md bg-purple-600"
-        onPress={handleLogout}
-      >
-        <Text className="pbk-h6 text-center text-base-white">Sign Out</Text>
-      </Pressable>
+      <Button
+        className="mb-4"
+        onPress={() => router.push("/(protected)/(draft)/applyAugment")}
```
Dennull: If you click this quickly the page loads twice. I'm not sure what the best way to solve this is. Maybe a throttle or disabling the button after the first press? Or turning it into a loading button? It's a bit weird, a throttle will probably work but technically it's a band-aid solution because if the navigation happens to take a long time for example the user could still click the button twice after the throttle wears off

client/components/Dialog.tsx
```diff
@@ -0,0 +1,41 @@
...
+  <Modal
+    animationType="fade"
+    onRequestClose={onClose}
+    transparent
+    visible={visible}
+  >
+    <View className="flex-1 items-center justify-center bg-black/70 px-6">
+      <View className="w-full rounded-xl bg-gray-920 p-6">
+        <View className="mb-4 flex-row items-center justify-between">
+          <Text className="pbk-b1 text-left text-gray-300">{title}</Text>
```
Dennull, line 30: Do you guys think the title be centered? @NasihNazeem @JayChuang8

client/components/Dialog.tsx
```diff
@@ -0,0 +1,41 @@
...
+            className="size-8 items-center justify-center rounded-md border border-gray-900"
+            onPress={onClose}
+          >
+            <X color="white" size={20} weight="bold" />
+          </Pressable>
+        </View>
+
+        <View className="mb-4">{children}</View>
+
+        <Button onPress={onClose} text="Got It" variant="secondary" />
```
Dennull: Maybe this confirm text should be a prop and "Got It" is the default in case we want to change it

client/app/(protected)/(draft)/applyAugment.tsx
```diff
@@ -0,0 +1,229 @@
...
+                <Text className="pbk-h5 text-base-white">
+                  Select one augment
+                </Text>
+                <HelpDialogButton title="What are Augments?">
+                  <Text className="pbk-b2 text-gray-500">
+                    {`Augments are strategic bonuses that reward how you build your team.\n\nEach Augment comes with a unique condition. When certain players on your roster meet that condition, only those players get a boost to their stats.\n\nWhether you're stacking sharpshooters, loading up on defenders, or betting on underdogs, Augments let you shape your playstyle and gain a competitive edge.`}
+                  </Text>
+                  <View className="mt-3 rounded-md bg-gray-900 p-3">
+                    <Text className="pbk-b2 text-base-white">
+                      {`Example\n\nThe “Deep Threat” Augment gives a +15% Points boost to players shooting 40%+ from three — but only if you have 3 such players on your team.`}
```
Dennull: Do you guys think "Example" be centered? Also, another nit but should we use css styling over \n\n for line breaks? @JayChuang8 @JayChuang8

client/controllers/userController.ts
```diff
@@ -122,7 +124,7 @@ export class UserController {
   static editUserTeam = async (
     userId: string,
     teamId: string,
-    params: TeamEditModel,
+    params: Partial<TeamEditModel>,
```
Dennull, line 130: Wait I'm not sure how it works for mongo, if this `params` parameter is partial and is missing some things does it only change whatever's present? For example if we're only passing `augmentId` and we're missing like `logoUrl` does `logoUrl` in the database stay the same? Or does it get set to null?

client/types/augmentTypes.ts
Dennull, line 1: More of a general comment, should we have the "Types" suffix on these files? @JayChuang8 @NasihNazeem

client/app/(protected)/(draft)/applyAugment.tsx
```diff
@@ -0,0 +1,229 @@
...
+const ApplyAugment = () => {
+  const userId = useAppSelector((state) => state.user.id);
+  const teamId = useAppSelector((state) => state.team.id);
+  const augmentId = useAppSelector((state) => state.team.augmentId);
+
+  const dispatch = useAppDispatch();
+
+  const [isLoading, setIsLoading] = useState(false);
+  const [selectedAugmentId, setSelectedAugmentId] = useState<string>(
+    augmentId || "",
```
Dennull: `augmentId` can be undefined in the team slice, should we either set this to undefined or update the type in the slice to not allow undefined?

client/app/(protected)/(draft)/applyAugment.tsx
Dennull, line 1: This looks really good, I like the gradient!

Dennull (changes requested):

client/controllers/userController.ts
JayChuang8: Yeah it just stays the same if you don't pass in parameters. It only updates what you pass in

client/assets/images/augments/background-image.png
JayChuang8: Yeah I mentioned it in the description, we're going to have to move all static images to a blob storage like S3 or google cloud storage. Since we're already using Firebase/Google maybe we use Firebase Cloud storage.

client/app/(protected)/(draft)/applyAugment.tsx
NasihNazeem: I'm actually thinking of the same thing in my current ticket. Since phosphor doesn't allow us to modify as we choose, I think it's better to wrap it with a view and just deal with the styling there. Maybe we can create an Icon component that takes in a phosphor icon, and has a prop for classname and sizing or whatever so its more readable and we don't have a bunch of View wrappers.

client/components/Dialog.tsx
NasihNazeem: I think aligned left is standard, I checked in with Nasrif and Jay offline as well

client/types/augmentTypes.ts
NasihNazeem: I'm indifferent tbh, I can understand that the assumption is there that it is a .ts file under a types folder, so we don't really need to specify that they are Types in the file name. But I'm cool with whatever if you guys have a preference.

client/app/(protected)/(draft)/applyAugment.tsx
JayChuang8: Yeah I just changed this useState to allow undefined

client/app/(protected)/(draft)/applyAugment.tsx
JayChuang8: It doesn't look too bad centered so I've centered it now. I also switched over to CSS, I previously had them with \n\n so I could use the template string and the apostrophe wouldn't be called out by eslint. I have it as `&apos;` now because of the CSS styling method

client/components/Button.tsx
JayChuang8: Yeah, changed to `label` and required now

client/types/augmentTypes.ts
JayChuang8: Ya since the folder is already called types I removed the types suffix now

client/app/(protected)/(draft)/applyAugment.tsx
JayChuang8: Yeah, I made a common `Screen` component that contains both the `SafeAreaView` and the keyboard dismiss `Pressable`

client/app/(protected)/(draft)/applyAugment.tsx
JayChuang8: I moved the opacity styling to the View component instead. For now, I’ll keep it as a View wrapping the Phosphor icon, since this is the only place in the app where we use a purely decorative, non-pressable icon. I don’t expect these purely cosmetic icons to be common, so I think this setup should be fine for now

client/app/(protected)/(tabs)/index.tsx
JayChuang8: Yeah I asked ChatGPT and the best way is to disable the button on press. I added a `isNavigating` state variable to handle this

client/app/(protected)/(draft)/applyAugment.tsx
Dennull: Oh nice that's great thanks

client/app/(protected)/(draft)/applyAugment.tsx
Dennull: Yeah that sounds good, if we need to have more of them we'll make a component

client/app/(protected)/(tabs)/index.tsx
Dennull: Yeah it feels kinda janky lmao but I don't have any better ideas tbh

client/types/augmentTypes.ts
Dennull: Ok, I like that

Dennull (approved):
I linked the augments page to the "build your team" button on the myTeam page. There are also some lingering uses of `Pressable`, but I'm working on some theming updates so I can update those to use our new `Button` component while I do that. It'll be a good chance to make all of our styling more consistent now that we have a solid theme and these wrappers. Nice work!

NasihNazeem (approved):

PR #16, opened 2025-08-18
Feature/ENG-9, ENG-44: My team layout, dynamic theming

Dennull:
## Related Issues

Closes ENG-9 and ENG-44

## Summary

Created the my team layout, set up nested tabs for the profile section, and added a dynamic theme that we can change based on user preference

## Changes Made

- Created a basic skeleton for the "My Team" page UI. There are already separate tickets for further functionality on this page
- Set up the file structure and routing for the nested tabs in the profile section
  - Also swapped completely to using the new `guard` prop for protected routes. Removed any remaining redirects
- Dynamic theming was set up, which allows for us to change the theme color/mode based on user preference
  - A provider and context was created so that you can access the currently selected theme and set the theme anywhere in the app. Just use the `useAppTheme()` hook
  - To use theme variables in tailwind, use prefixes as normal and use `primary-[number]`, `mode` or `modeContrast` as the value. For example, if you wanted to make the background the user's currently selected them you could do `className="bg-primary-600"`
  - Theme variables are set up in tailwind using css variables in the tailwind config. A plain JS object, `themes`, which stores all the theme colors as rgb values was also created in `theme.ts`. This is used in places that need a direct string value for colors and not a css variable, such as when updating the theme for `react-navigation`' (for our tabs)
    - The tailwind theme is automatically generated using this `theme` object along with the `applyCssTheme` helper function, so any theme updates should be done on the plain JS object

## Screenshots

https://github.com/user-attachments/assets/e702eaef-d2b2-4b36-97e6-785e253c62a5

## Testing Instructions

No additional notes


client/app/(protected)/(tabs)/(profile)/_layout.tsx
```diff
@@ -0,0 +1,59 @@
...
+      tabBarPosition: "top",
+      tabBarIconStyle: {
+        flex: 1,
+      },
+    }}
+  >
+    <Tabs.Screen
+      name="index"
+      options={{
+        title: "Roaster",
```
NasihNazeem: small little typo here for roster screen.

client/app/(protected)/(tabs)/(profile)/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+const Roaster = () => {
+  const team = useAppSelector((state) => state.team);
+  const matchedLogo = teamLogoOptions.find(
+    (option) => option.url === team.logoUrl,
+  );
+
+  return (
+    <View className="h-full w-full items-center justify-center bg-gray-950">
+      <View className="h-[250px] w-full">
```
NasihNazeem: I feel like we could get away with using `h-64` here right? Since I still consider this a layout container

client/app/(protected)/(tabs)/(profile)/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <View className="h-1/2 bg-pink-700" />
+        <View className="h-1/2 border-b border-gray-900 bg-gray-950">
+          <View className="-my-[37.5px] w-full flex-row justify-between px-8">
```
NasihNazeem: What do you think about using `-my-9` instead of `-my-[37.5px]`

client/app/(auth)/_layout.tsx
```diff
@@ -1,19 +1,23 @@
...
-  >
-    <Stack.Screen name="index" />
-    <Stack.Screen name="login" />
-    <Stack.Screen name="signUp" />
-    <Stack.Screen name="forgotPassword" />
-    <Stack.Screen name="confirmReset" />
-    <Stack.Screen name="emailVerification" />
-  </Stack>
+  <>
+    <StatusBar style="light" />
```
NasihNazeem, line 6: Just so I understand this is only light mode because it can't really be dark mode in the unprotected states right

client/app/(protected)/(tabs)/_layout.tsx
```diff
@@ -1,52 +1,31 @@
...
+      options={{
+        title: "Home",
+        tabBarIcon: ({ color }) => <House color={color} />,
+      }}
+    />
+    <Tabs.Screen
+      name="(profile)"
+      options={{
+        title: "Profile",
+        tabBarIcon: ({ color }) => <IdentificationBadge color={color} />,
```
NasihNazeem: Are we calling this folder profile? I feel like that could be confusing with the other part of our app that includes the icon. I believe @nasrif  is looking into making changes to the tab bar anyway so maybe he can pitch in on what that tab should be called. 

I've included a screenshot of the tab bar I am talking about with the other "profile" icon
{D0887A9C-4BD7-435A-9679-6AF41B2C10D7}

client/theme/ThemeProvider.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+const ThemeContext = createContext<ThemeContextType | undefined>(undefined);
+
+type ThemeProviderProps = {
+  children: ReactNode;
+};
+
+export const ThemeProvider = ({ children }: ThemeProviderProps) => {
+  const [theme, setTheme] = useState<ThemeName>(ThemeName.Purple);
+  const [mode, _setMode] = useState<ThemeMode>(ThemeMode.System);
```
NasihNazeem, line 20: These two things should be stored in AsyncStorage I think

We would have to create like a setPersistentTheme func that sends a key for both theme and mode async using `AsyncStorage.setItem()`. 

And probably need a loadTheme() func under a useEffect that gets the saved values.

Let me know if this makes sense

client/types/avatars.ts
NasihNazeem, line 1: If we really wanted to, you could combine avatars and teamLogo type into the same thing. Maybe call the type Asset or something along those lines?

client/app/(protected)/(tabs)/(profile)/matchHistory.tsx
NasihNazeem, line 1: I believe this is a post-MVP feature now confirmed by @nasrif under ENG-40 issue. 

https://linear.app/rookies-fantasy/issue/ENG-40/viewing-match-history#comment-0da7d4d6

client/theme/theme.ts
NasihNazeem, line 1: Do we include semantic colors like success, warning, errors, info in here?

client/app/_layout.tsx
```diff
@@ -11,33 +16,51 @@ import { SafeAreaProvider } from "react-native-safe-area-context";
...
+
+  const testNavTheme = {
+    ...(mode === ThemeMode.Light ? DefaultTheme : DarkTheme),
+    colors: {
+      ...(mode === ThemeMode.Light ? DefaultTheme.colors : DarkTheme.colors),
+      primary: `rgb(${themes[theme][mode][500]})`,
+      background: `rgb(${themes[theme][mode].mode})`,
+      text: `rgb(${themes[theme][mode].modeContrast})`,
+    },
+  };
```
NasihNazeem, line 44: This looks like a temp name but I think we still need this so we should call it `navigationTheme`

NasihNazeem (changes requested):
This is great progress, you basically created a whole theming library here. 

I believe most of my comments were small feedback or questions but one was about setting theme in async storage. Let me know if you have any questions.

client/app/(protected)/(tabs)/_layout.tsx
nasrif: Profile is the one on the far right

client/app/(protected)/(tabs)/(profile)/matchHistory.tsx
NasihNazeem: As discussed offline, we will keep this page as this version of the application still requires this page.

client/app/(protected)/(tabs)/(profile)/_layout.tsx
Dennull: Oops lol

client/app/(protected)/(tabs)/(profile)/index.tsx
Dennull: I left it as fixed pixel width rather than rems because I found that other apps have profile sections which also do the same, I think they have banners which are the same size no matter what device you're on. As discussed offline we don't feel that strongly about it tho so we can maybe explore changing it later

client/app/(protected)/(tabs)/(profile)/index.tsx
Dennull: Similar comment to height one on line 13

client/app/(auth)/_layout.tsx
Dennull: Yes that's correct in the auth routes we don't let the user set their modes so it'll always be dark mode

client/app/(protected)/(tabs)/_layout.tsx
Dennull: As discussed offline these changes are for future iterations. For now, the highlighted tab there is called "My Team", and the first subnav (where you view your team name and can click on the build a team button) is called "Roster"

client/theme/ThemeProvider.tsx
Dennull: Yes this does make sense, but I'd rather explore that in another ticket if that's ok

client/app/_layout.tsx
Dennull: Oops that's embarrassing lol

client/theme/theme.ts
Dennull: Yeah I thought about this, but as discussed offline, the JS object version of our theme is only used for specific instances where we need to access the raw string (ie. hex or rgb values) instead of the css variable version. I don't think we'll be using the string values of our semantic colors very often so let's just leave them out for now

You can still use those colors in tailwind though since those colors are included in the tailwind theme; just use something like `className="bg-red-500"` as we have been doing for example. Of course it's a relatively easy fix to add them in our JS object later if we need as well

client/types/avatars.ts
Dennull: Yeah sure that makes sense to me, it's the same type basically

JayChuang8 (approved):

NasihNazeem (approved):

PR #18, opened 2025-09-04
Feature/ENG-25: Add table filters for player selection

NasihNazeem:
## Related Issues

Closes ENG-25. 

## Summary

Implements a comprehensive filtering system for the player table in the team builder, allowing users to filter players by team, position, and salary range.

NOTE: Review this after ENG-32 and ENG-38 have been merged in.

## Changes Made

- Filter by team by choosing a specific NBA team with logo display
- Filter by positions (PG, SG, SF, PF, C) and flex positions (G, F, ALL)
- Filter by salary range using interactive slider (1M - 150M)
- Apply, reset and track filter counts

Custom Components:
- `FiltersDrawer` - Reusable drawer component that holds all the player filters
- `Accordion` - Reusable collapsible section component
- `RangeSlider` - Custom dual-thumb slider for salary range selection
    - I couldn't find any maintained library that gave me customizability and dual-thumbs
- Updated player table to integrate with filtering system
- Added a controller to handle querying for filtered players

## Screenshots

https://github.com/user-attachments/assets/9a2eb121-86c3-47bc-a8d6-860348e8cf39

## Testing Instructions

No additional notes


package-lock.json
Dennull, line 1: This is probably a mistake, when you run `npm i` in the wrong directory this happens lol

client/state/slices/teamSlice.ts
```diff
@@ -348,6 +348,11 @@ export const selectRosterPlayerCount = (state: RootState): number => {
   return lineupCount + benchCount;
 };
 
+export const isPlayerInLineup = (
```
Dennull: We moved this to `teamUtils.ts`

client/app/(protected)/(draft)/(teamBuilder)/players.tsx
```diff
@@ -20,33 +21,63 @@ import Table from "@/components/Table/Table";
...
+    salaryRange: { min: 1000000, max: 150000000 },
+  });
   const team = useAppSelector((state) => state.team);
   const userId = useAppSelector((state) => state.user.id);
   const dispatch = useAppDispatch();
   const router = useRouter();
 
+  const [showFiltersDrawer, setShowFiltersDrawer] = useState(false);
+
+  const activeFilters =
```
Dennull, line 57: This isn't a blocker, but do you think the filter count should be on individual filters selected or categories selected? I have seen apps only do categories selected

client/components/Accordion.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { ReactNode, useState, useEffect } from "react";
+import { Pressable, Text, View } from "react-native";
+import Animated, {
+  useAnimatedStyle,
+  useSharedValue,
+  withTiming,
+} from "react-native-reanimated";
+import { cn } from "@/utils/jsUtils";
+
+export type AccordionItem = {
```
Dennull: This isn't used. Also, if we're exporting types let's try to make new type files for them

client/components/Accordion.tsx
```diff
@@ -0,0 +1,106 @@
...
+  id: string;
+  title: string;
+  content: ReactNode;
+  disabled?: boolean;
+  defaultOpen?: boolean;
+};
+
+export type AccordionProps = {
```
Dennull: We don't need to export this

client/components/Accordion.tsx
```diff
@@ -0,0 +1,106 @@
...
+  animationDuration?: number;
+};
+
+const Accordion = ({
+  title,
+  children,
+  defaultExpanded = false,
+  disabled = false,
+  onChange,
+  isExpanded: controlledIsExpanded,
```
Dennull: Do you think we need this? It gets a little confusing and naming feels clunky when we have this controlled prop. I sort of doubt we'll ever need to control the expanded state of the accordion externally in the first place

Since this is a component only we're using I feel like we can ditch this for now and build it in later if we need. It would reduce the boilerplate on lines 45 and 56 down too

client/components/Accordion.tsx
```diff
@@ -0,0 +1,106 @@
...
+  disabled?: boolean;
+  defaultOpen?: boolean;
+};
+
+export type AccordionProps = {
+  title: string;
+  children: ReactNode;
+  defaultExpanded?: boolean;
+  disabled?: boolean;
+  onChange?: (isExpanded: boolean) => void;
```
Dennull: Do you think `onToggle` fits this better?

client/components/Accordion.tsx
```diff
@@ -0,0 +1,106 @@
...
+  contentClassName,
+  animationDuration = 200,
+}: AccordionProps) => {
+  const [localIsExpanded, setLocalIsExpanded] = useState(defaultExpanded);
+  const isControlled = controlledIsExpanded !== undefined;
+  const isExpanded = isControlled ? controlledIsExpanded : localIsExpanded;
+  const rotation = useSharedValue(isExpanded ? 90 : 0);
+  const height = useSharedValue(isExpanded ? 1 : 0);
+
+  useEffect(() => {
```
Dennull, line 40: I think the animation on this could be improved. It doesn't look super smooth right now. However, that's something we can do later. Good job getting it working right now. Just wanted to note it down. I'll make a ticket so it's on our radar

client/components/BottomSheet.tsx
```diff
@@ -13,19 +13,19 @@ import Animated, {
 } from "react-native-reanimated";
 import { SafeAreaView } from "react-native-safe-area-context";
 
-type SnapPoint = "0%" | "33%" | "50%" | "66%" | "100%";
-
 export const { height: SCREEN_HEIGHT } = Dimensions.get("window");
-export const MAX_TRANSLATE_Y = -SCREEN_HEIGHT + 50;
+export const MAX_TRANSLATE_Y = -SCREEN_HEIGHT + 150;
```
Dennull: What's this magic number for? Do you think we could make a variable or add a comment to document it?

client/components/RangeSlider.tsx
```diff
@@ -0,0 +1,319 @@
...
+import { Gesture, GestureDetector } from "react-native-gesture-handler";
+import Animated, {
+  useAnimatedStyle,
+  useSharedValue,
+  runOnJS,
+  withTiming,
+} from "react-native-reanimated";
+import { cn } from "@/utils/jsUtils";
+
+export type RangeSliderValue = [number, number];
```
Dennull: Same here, we don't need to export these types

client/components/RangeSlider.tsx
Dennull, line 1: I have a few thoughts about this component, but none of them really worth sharing in this PR specifically. We're short on time and this works as it is so I'll chat with you about them later or open a separate PR. The general topic though is that I feel like there are almost too many bells and whistles with this component - customization is good, but having a ton of props and ways to customize can get hard to use and easy to break, so I'd like to consider slimming this component down a bit

client/controllers/nbaPlayersController.ts
```diff
@@ -68,4 +69,101 @@ export class NbaPlayersController {
       throw error;
     }
   };
+
+  static getFilteredPlayers = async (
```
Dennull, line 109: Ideally I'd love to rework this to use the same endpoint as our `getPlayers` endpoint. We could probably find a way to build the query more dynamically and then return all players if you have no filters. But this already works, so if that's a bigger task I'd suggest making a new ticket to document it and changing that later

client/types/team.ts
```diff
@@ -27,7 +28,16 @@ export type BenchSlot = {
   position: BenchPosition;
   player: Player;
 };
-
+export const POSITION_OPTIONS: PositionOption[] = [
+  "ALL",
```
Dennull: Do we need this? If you have none selected it returns all right, so this is sort of redundant. It leads to weird cases too like if I have "ALL" selected with any other thing those other filters get overwritten

client/types/team.ts
```diff
@@ -17,6 +17,7 @@ export type Position = "PG" | "SG" | "SF" | "PF" | "C";
 export type FlexPosition = "UTIL1" | "UTIL2" | "UTIL3";
 export type BenchPosition = `BEN${number}`;
 export type SlotPosition = Position | FlexPosition | BenchPosition;
+export type PositionOption = Position | "ALL" | "G" | "F";
```
Dennull: Nit, could we name this PositionFilters? Same with the variable on line 31. These are technically filter options, not real player positions

client/app/(protected)/(draft)/(teamBuilder)/players.tsx
```diff
@@ -20,33 +21,63 @@ import Table from "@/components/Table/Table";
...
 import { Player } from "@/types/players";
-import { FlexPosition, UTIL_POSITIONS } from "@/types/team";
+import { FlexPosition, PositionOption, UTIL_POSITIONS } from "@/types/team";
 import { isPlayerInLineup, resetTeamLineup } from "@/utils/teamUtils";
 
 type FetchPlayersParams = {
   pageParam?: FirebaseFirestoreTypes.DocumentSnapshot;
 };
 
+export type Filters = {
```
Dennull: This sort of feels weird to be exported here; this filtering behavior is pretty generic and could be applied to any player table. I'd also like to rename this to be more specific, maybe like PlayerFilters. Since we have a few more player related types, we should make a `players.ts` type/constant file and put this there, what do you think?

client/components/FiltersDrawer.tsx
Dennull, line 1: A small UX improvement could be made - when all of the accordions are collapsed there's no indication of which categories have a filter. It'd be nice if we could find a way to show that. Maybe we add a little label like "Team (2 selected)". We could also add a badge. But we already have a badge on the icon button, so idk. A few options to explore

Dennull (changes requested):

client/app/(protected)/(draft)/(teamBuilder)/players.tsx
NasihNazeem: I agree with the suggestion, I can add it to the players.ts file we currently have

client/app/(protected)/(draft)/(teamBuilder)/players.tsx
NasihNazeem: I guess it is more intuitive to be thinking "I have 2 filters applied" or "I am filtering by team and players, rather than saying I am filtering by these two teams and these two positions".

I will make the adjustment to count for filters being applied, similar to what I used with the salary range.


```suggestion
    const activeFilters =
        (filters.selectedTeams.length > 0 ? 1 : 0) +
        (filters.selectedPositions.length > 0 ? 1 : 0) +
        (filters.salaryRange.min !== 1000000 ||
        filters.salaryRange.max !== 150000000
            ? 1
            : 0);
```

client/components/Accordion.tsx
NasihNazeem: Okay sounds good

client/components/Accordion.tsx
NasihNazeem: Good catch, thank you

client/components/Accordion.tsx
NasihNazeem: I was thinking if a parent component ever needed to know the state of the expansion it could be helpful but I didn't really use it in this PR. Tbh I can say that for almost all my props in this component aside from `title` and `children` .

Since I don't actually use it I can remove this for now. How do you feel about me removing some of the other props that I don't use right now?

NasihNazeem: I guess in practice it wouldn't make sense for a parent to need the `isExpanded` state, `onToggle` should suffice for future cases.

client/components/BottomSheet.tsx
NasihNazeem: I don't actually have a good reason here. I think I just found that 90 wasn't a true representation of 90 so I thought adding a offsetting by a 100 would look better.

On second thought, I think adjusting the snap point makes more sense so I am going to use an 80% snap point instead of a 90%

client/types/team.ts
NasihNazeem: I understand, I thought so too initially but I went with it to match with the design provided since I didn't really bring it up in a meeting to talk about changing it. I will make the adjustment now since it doesn't take too much.

client/state/slices/teamSlice.ts
NasihNazeem: Good catch, this didn't seem to pop up in the conflicts

client/types/team.ts
NasihNazeem: Good point

client/controllers/nbaPlayersController.ts
NasihNazeem: I think you're right. I think a simple adjustment would be to introduce an optional filters argument in the original and then have an if statement handling if the query gets adjusted.

client/components/RangeSlider.tsx
NasihNazeem: I can see your point. My line of thinking was if i'm going to make a custom slider, I should allow some more freedom with how it works and looks.

To simplify this, I can maybe reduce the props to the essentials `min`, `max`, `step`, `formatValue`, `onChange`, `value` and keep the rest to what I have as a default. What do you think?

package-lock.json
NasihNazeem: Yeah it is, whoops lol

client/state/slices/teamSlice.ts
NasihNazeem: Good catch, thank you

client/components/BottomSheet.tsx
NasihNazeem: I wasn't sure if the + 50 was accurate to get my 90% screen height, but on second thought I think I just needed to adjust the snap point to be 80% rather than 90%.

client/components/RangeSlider.tsx
NasihNazeem: Yessir, this doesn't need to be exported.

client/components/FiltersDrawer.tsx
NasihNazeem: I can add a label stating "2 selected", I think that makes sense. Thanks for the suggestion.

client/app/(protected)/(draft)/(teamBuilder)/players.tsx
Dennull: Yeah, I like it!

client/components/Accordion.tsx
Dennull: I could see uses for the other props, so I don't mind leaving them. In general though I'd lean towards keeping things simpler at first and then adding props down the line if we ever need them 👍

client/components/Accordion.tsx
Dennull: https://linear.app/rookies-fantasy/issue/ENG-39/smooth-out-accordion-animation

client/components/RangeSlider.tsx
Dennull: Yeah, I'd lean towards that. But for now I think we can leave it, it works fine and looks good. I had some ideas so maybe we can go through it together one day

client/controllers/nbaPlayersController.ts
Dennull: Yeah, I think that's the better way. Here's a ticket to track it https://linear.app/rookies-fantasy/issue/ENG-50/refactor-getplayers-endpoint-to-take-filter-parameters

client/app/(protected)/(draft)/(teamBuilder)/players.tsx
Dennull: Oh yeah nice I forgot we already had one, thanks!

client/components/FiltersDrawer.tsx
Dennull: Wow nice job it looks great!!

client/controllers/nbaPlayersController.ts
NasihNazeem: I attempted this fix in this PR actually, lmk if it works for you

Dennull: Oh great! Thanks for doing that

Dennull (approved):
Nice job!

PR #19, opened 2025-09-18
Feature/ENG-47: Update augmentId in teams collection to use augment object instead

NasihNazeem:
## Related Issues

Closes ENG-47.

## Summary

This PR is to update the use of augmentId to now just store the entire object instead. This will ease the validation flow as we do not need to fetch and store the selected augment separately, and can simply store it within the team state.

## Changes Made

- Update team state to use entire Augment rather than just augmentId
- Fetch for prerequisites, effects and isActive boolean on all augments
- Update `selectAugment` reducer to set entire augment
- Update `applyAugment.tsx` to select entire augment instead of augmentId, and update functions to reflect type changes.

## Screenshots

No additional screenshots

## Testing Instructions

If you would like to, please remove the augmentId stored in your teams subcollection as we will be storing the entire augment there now.

No additional notes

### DIFF
--- a/client/types/augment.ts
+++ b/client/types/augment.ts
@@ -4,4 +4,85 @@ export type Augment = {
   id: string;
   info: string;
   title: string;
+  isActive: boolean;
+  updatedAt?: string;
+  createdAt?: string;
+  prerequisites: Prerequisite[];
+  effects: Effect[];
+};
+
+export type Condition = {
+  count: number;
+  stat?:
+    | "points"
+    | "rebounds"
+    | "assists"
+    | "steals"
+    | "blocks"
+    | "threePointers"
+    | "fieldGoalPercentage"
+    | "freeThrowPercentage"
+    | "turnovers"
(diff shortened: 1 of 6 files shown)


client/types/augment.ts
```diff
@@ -1,7 +1,84 @@
...
   info: string;
   title: string;
+  isActive: boolean;
+  updatedAt?: string;
+  createdAt?: string;
+  prerequisites: Prerequisite[];
+  effects: Effect[];
+};
+
+type Stat =
```
Dennull, line 16: Is this stuff for the future augment PR?

Dennull (approved):

client/types/augment.ts
NasihNazeem: Yep, but if you're talking about this type specifically, it's just so that I don't have a bunch of or operators all over

I've got the other branches ready, I just wanted to do them one at a time

PR #20, opened 2025-09-22
Feature/ENG-38: Team slice refactors

Dennull:
## Related Issues

Related to ENG-38

## Summary

Made some further refactors to the team slice and roster drawer. The original PR was getting bloated already, so I opted to make a separate one. Mainly making things more clear and removing repetitive code.

## Changes Made

- Added a few more helper functions and used them in the roster drawer and the team slice. Also consolidated some old helper functions all into one
- More refactors to the `swapPlayersInLineup` reducer. That logic was pretty complicated and hard to read, so I slimmed things down by encapsulating code that was repeated into functions. Also renamed some variables to read better
- I noticed that the `addPlayerToLineup` reducer had some code that was repetitive too so I removed some redundancy

## Screenshots

No additional screenshots

## Testing Instructions

No additional notes

### DIFF
--- a/client/state/slices/teamSlice.ts
+++ b/client/state/slices/teamSlice.ts
@@ -130,7 +130,7 @@ const teamSlice = createSlice({
 
       for (const position of allEligiblePositions as SlotPosition[]) {
         const slot = findSlotFromPosition(state.lineup, position);
-        if (isNotNil(slot) && !isNotNil(slot.player)) {
+        if (isNotNil(slot) && isNil(slot.player)) {
           slot.player = player;
           state.balance -= slot.player.salary;
           state.hasUserChanges = true;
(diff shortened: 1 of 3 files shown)


NasihNazeem (approved):
No blockers, we can merge

client/state/slices/teamSlice.ts
```diff
@@ -11,71 +11,99 @@ import {
...
-
-const removeBenchSlot = (bench: BenchSlot[], position: BenchPosition) => {
-  const index = bench.findIndex((slot) => slot.position === position);
-  if (index !== -1) {
-    bench.splice(index, 1);
-    bench.forEach((slot, i) => {
-      slot.position = `BEN${i + 1}` as BenchPosition;
+const swapBenchAndLineup = (
+  lineupSlot: LineupSlot,
+  benchSlot: BenchSlot,
```
NasihNazeem, line 82: do you think we should specify incoming similar to what you did when passing props

Dennull: Nah, you don't know which is incoming and current in this case because it's between a lineup and bench slot right. Like I could be moving someone from the bench to the lineup or the reverse

NasihNazeem: Oh ok I misunderstood, thank you for explaining

PR #26, opened 2025-11-25
Feature/ENG-58: Fetch live data from API

Dennull:
## Related Issues

Addresses ENG-58 and ENG-64.

## Summary

Update the arena page with live basketball stats. We fetch for new stats every so often and update the screen with them

## Changes Made

- Made `getLiveData`, a cloud function that can be called via HTTP to fetch the latest game data. The function keeps a cache in RAM of the latest player stats and serves that if available
  - If someone hits the endpoint and the cache is older than 60 seconds, it refetches from the balldontlie API
- On the front end, we implement polling on the arena page. It polls `getLiveData` every 10 seconds. We should remove this in the future and find a better way to do it, but for now this should work

## Screenshots

No additional screenshots

## Testing Instructions

No additional notes


Dennull:
<p><a href="https://linear.app/rookies-fantasy/issue/ENG-58/setting-up-redis">ENG-58 Setting up Redis</a></p>

client/app/(protected)/(tabs)/(home)/arena.tsx
```diff
@@ -1,18 +1,27 @@
...
-import { selectMatchup } from "@/state/slices/matchupSlice";
+import { useAppDispatch, useAppSelector } from "@/state/hooks";
+import {
+  selectMatchup,
+  updateMatchupWithLiveData,
+} from "@/state/slices/matchupSlice";
 import { teamLogoOptions } from "@/types/asset";
 import { SLOT_ORDER } from "@/types/team";
 
+const LIVE_DATA_URL =
```
NasihNazeem, line 19: What do you think about storing the url, endpoint aside into an env?

Maybe we make a ticket on this

client/app/(protected)/(tabs)/(home)/arena.tsx
```diff
@@ -32,12 +41,80 @@ const Arena = () => {
...
+            }),
+            fetch(LIVE_DATA_URL, {
+              method: "POST",
+              headers: {
+                "Content-Type": "application/json",
+                Authorization: `Bearer ${idToken}`,
+              },
+              body: JSON.stringify({ playerIds: homePlayerIds }),
+            }),
+          ]);
```
NasihNazeem, line 88: Let's make a ticket on this to batch the data into one api call

client/app/(protected)/(tabs)/(home)/arena.tsx
```diff
@@ -32,12 +41,80 @@ const Arena = () => {
...
+              updatedHome,
+              updatedAway,
+            }),
+          );
+        } catch (error) {
+          console.log(error);
+        }
+      };
+
+      fetchLiveData();
```
NasihNazeem, line 107: This might require tweaking the matchup structure again, but it would be nice to make it so that this only gets called if there is a live game ongoing for atleast 1 player in the matchup.

Maybe we make a ticket to see investigate how we can optimize the polling rather than poll always

client/app/(protected)/(tabs)/(home)/arena.tsx
```diff
@@ -32,12 +41,80 @@ const Arena = () => {
...
+          dispatch(
+            updateMatchupWithLiveData({
+              date: currentDate,
+              updatedHome,
+              updatedAway,
+            }),
+          );
+        } catch (error) {
+          console.log(error);
+        }
```
NasihNazeem, line 104: Related but not only for this ticket, we need to solve the error handling all over the app and how we communicate that to the user. This is an important issue to deal with, but I'm not sure if we will get to it this sprint.

client/state/slices/matchupSlice.ts
```diff
@@ -1,13 +1,52 @@
...
+          } | null
+        >;
+        updatedAway: Record<
+          string,
+          {
+            gameInfo: GameInfo;
+            gameStats: GameStats;
+          } | null
+        >;
+      }>,
```
NasihNazeem, line 24: In-line type definition is a bit hard to read, I think we can push this elsewhere in the file so that the reducer is easier to understand. Let me know what you think

client/state/slices/matchupSlice.ts
```diff
@@ -1,13 +1,52 @@
...
+          o.gameStats = newData?.gameStats;
+        }
+      });
+      newMatchup.awayTeam.lineup.forEach((o) => {
+        if (isNotNil(o.player?.id)) {
+          const newData = action.payload.updatedAway[o.player?.id];
+          o.gameInfo = newData?.gameInfo;
+          o.gameStats = newData?.gameStats;
+        }
+      });
```
NasihNazeem, line 40: We can make a ticket to maybe make a utility function for this so we reduce duplicated code

server/functions/src/index.ts
```diff
@@ -1,9 +1,148 @@
...
+            tov: player.turnover ?? 0,
+            fpts: player.fpts ?? 0,
+            min: parseInt(player.min, 10) || 0,
+          },
+        };
+      }
+
+      for (const player of awayTeam.players) {
+        const gameInfo: GameInfo = {
+          gameStatus: game.status !== "Final" ? false : true,
```
NasihNazeem, line 116: I wanted to change this in my ticket, but I didn't want to mess with it. We should change how we apply `gameStatus`

server/functions/src/index.ts
```diff
@@ -1,9 +1,148 @@
...
+  number,
+  {
+    gameInfo: GameInfo;
+    gameStats: GameStats;
+  }
+>;
+
+const apiKey = process.env.BALLDONTLIE_API_KEY || "";
+const api = new BalldontlieAPI({ apiKey });
+const cachedData: LiveData = {};
```
NasihNazeem, line 36: Caching in the cloud function works as a band-aid because I think if Firebase is using multiple instances of this function, each instance will have its own dedicated cache which doesn't have the context of another cache.

We should definitely look to move past this, but I'm not sure if this is breaking enough of the app for us to worry about it this sprint.

server/functions/src/index.ts
```diff
@@ -1,9 +1,148 @@
...
+            reb: player.reb ?? 0,
+            stl: player.stl ?? 0,
+            blk: player.blk ?? 0,
+            tov: player.turnover ?? 0,
+            fpts: player.fpts ?? 0,
+            min: parseInt(player.min, 10) || 0,
+          },
+        };
+      }
+    }
```
NasihNazeem, line 135: We can reduce duplicated code here by creating a process function

NasihNazeem (approved):
I'm going to approve these changes but I did leave some comments for potential new tickets for us to work on. Let me know your thoughts, I didn't want to make the tickets right away unless someone else agrees to move forward with the proposed tickets

client/state/slices/matchupSlice.ts
JayChuang8: Yeah i agree, maybe extract the `GameInfo` and `GameStats` into a type

JayChuang8 (approved):

client/app/(protected)/(tabs)/(home)/arena.tsx
Dennull: We can if we want. I don't feel strongly either way because I think this is a public endpoint anyways. Anyone can craft a web request and hit it, we're not exposing any secrets

client/app/(protected)/(tabs)/(home)/arena.tsx
Dennull: Yeah the way this is set up right now it's kind of hard to do that, but we can certainly look into it. I made a ticket for a lot of these optimizations https://linear.app/rookies-fantasy/issue/ENG-88/optimize-live-data-fetching

client/app/(protected)/(tabs)/(home)/arena.tsx
Dennull: Yup, added it to [ENG-88](https://linear.app/rookies-fantasy/issue/ENG-88/optimize-live-data-fetching)

client/app/(protected)/(tabs)/(home)/arena.tsx
Dennull: Yeah lol I think we had a really old ticket for this already https://linear.app/rookies-fantasy/issue/ENG-18/establish-error-handling-patterns. It's just not a top priority right now when we're trying to get a proof of concept

client/state/slices/matchupSlice.ts
Dennull: Yeah, I at least extracted that object into a type in this slice. We could also change that to be the same elsewhere because the `LineupSlot` type has both of these too, but I wasn't sure if we wanted to do that yet

client/state/slices/matchupSlice.ts
Dennull: I think this should be done as a part of [ENG-65](https://linear.app/rookies-fantasy/issue/ENG-65/restructure-matchup-state-and-collection-structure-to-reduce-the)

server/functions/src/index.ts
Dennull: Yeah, it's kind of dumb that their API is like this. The return structure is kind of shitty too. I think we can tackle those things as part of [ENG-81](https://linear.app/rookies-fantasy/issue/ENG-81/refactor-typing-on-server-side-cloud-functions)

server/functions/src/index.ts
Dennull: Yeah Redis is definitely the tool to use for something like this. I had kind of a duplicate ticket of this floating around so I repurposed it to be for setting up Redis https://linear.app/rookies-fantasy/issue/ENG-64/set-up-redis-cache-for-live-game-data

server/functions/src/index.ts
Dennull: Word, I added this to [ENG-88](https://linear.app/rookies-fantasy/issue/ENG-88/optimize-live-data-fetching)

PR #29, opened 2025-11-25
Feature/ENG-57: Create arena page

NasihNazeem:
## Related Issues

Closes ENG-57.

## Summary

This PR introduces the Arena, where users can view their matchup from a daily team vs team perspective. This includes player-player comparisons, live scoring, augment displays, and date navigation. This includes significant refactoring of the matchup data structure on both client and server to support richer matchup functionality.

## Changes Made

- Create `PlayerMatchupCard`
    - Head-to-head player comparisons for home/away players in particular positions.
- Create `DateSelector`
    - Week-long date navigation
    - Disabled state at boundaries
- Update `Arena` Page
    - Complete matchup viewing interface with team metadata, team players, and scores
    - Split design "Home vs Away"
    - Interactive augment display cards with dialog modals
    - Scrollable player matchup list ordered by position
    - Date selector for navigating through the matchup days
- Create `matchupSlice.tsx`
    - New slice to handle setting matchups
- Create `matchup.ts` type definitions file
- Update `team.ts` type defintions introducing matchup state (GameInfo, GameStats)
- Create `matchupController.ts`
- Real-time updates in `AuthListener`
    - Wait for queue status to be matched, and once matched, trigger a matchup fetch to fill in data from firestore
- Race condition fix for matchups
    - Introduce `matching` state to prevent duplicate matchup creation
    - Restructred matchup document
    - Game Data integration
- Removed `queueController.ts` as we consolidated the functionality in other controller
- Fixed some of the spacing and alignment issues
- Added disabled prop to `IconButton`
 
**Matchup Data Structure:**
```
Matchup {
  id, status, weekStartDate,
  home: { homeUserId, homeTeamId, homeTeamName, homeTeamLogo, homeAugment, ... },
  away: { awayUserId, awayTeamId, awayTeamName, awayTeamLogo, awayAugment, ... },
  dailyMatchups: {
    "2024-11-24": { homeTeam: {...}, awayTeam: {...} },
    "2024-11-25": { homeTeam: {...}, awayTeam: {...} },
    ...
  }
}
```

## Screenshots

No additional screenshots

## Testing Instructions

- [ ] Verify arena page loads with correct team data
- [ ] Test date selector navigation
- [ ] Confirm player cards display correctly
- [ ] Validate augment modal 
- [ ] Test matchup data loading on auth
- [ ] Verify race conditions mentioned in changes made
- [ ] Confirm monday next week start calculation is correct
- [ ] Test game info fetching
- [ ] Validate lineup slots have game metadata
- [ ] Check matchup state clears on logout
- [ ] Test with incomplete lineups
- [ ] Verify scoring displays correctly

No additional notes


NasihNazeem:
<p><a href="https://linear.app/rookies-fantasy/issue/ENG-57/arena-skeleton-layout">ENG-57 Arena skeleton layout</a></p>

server/functions/src/index.ts
```diff
@@ -67,39 +65,67 @@ export const processQueue = functions
...
               });
-
-              console.log(
-                `Match created: ${matchupId} between ${user1.id} and ${user2.id}`,
-              );
             }
           }
         });
 
+        const waitingUsers = await admin
```
Dennull, line 81: Is there a reason this isn't happening in the transaction block above? It feels weird that we're querying for the same users again when we already found 2 users we want to match up. As a result the "matching" status seems a bit redundant too

server/functions/src/index.ts
Dennull, line 1: The way this is set up for creating the matchup was slightly hard to follow for me. I think I get how it works now though, but maybe it could be improved. Part of it is also probably because there's no typing. I'm gonna have to revisit this when I work on the weekly acquisitions, so if you're down do you wanna you hop on a call to review with me then and see if we can improve it?

client/controllers/matchupController.ts
```diff
@@ -0,0 +1,64 @@
...
+      away: data.away || {},
+      home: data.home || {},
+      status: data.status ?? "active",
+      weekStartDate: data.weekStartDate,
+      dailyMatchups: {},
+    };
+
+    if (data.weekStartDate) {
+      matchup.weekStartDate = data.weekStartDate;
+    }
```
Dennull: What happens if this is missing? Would that ever happen?

client/controllers/matchupController.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    Object.keys(data).forEach((key) => {
+      if (/^\d{4}-\d{2}-\d{2}$/.test(key)) {
```
Dennull, line 48: This is to check if this is a valid date right? It seems like a weird thing to have to do and this way of checking also seems strange. I don't think it's worth worrying about now but it seemed weird to me so I wanted to comment on it. If you also agree let's leave a TODO, otherwise it's whatever

client/components/PlayerMatchupCard.tsx
Dennull, line 1: I feel like we can refactor this to be a bit cleaner but let's leave it for another day https://linear.app/rookies-fantasy/issue/ENG-82/refactor-playermatchupcard-component

client/components/DateSelector.tsx
```diff
@@ -0,0 +1,76 @@
...
+    }
+  };
+
+  const handleNext = () => {
+    if (currentIndex < dates.length - 1) {
+      onDateChange(dates[currentIndex + 1]);
+    }
+  };
+
+  const formatDate = (dateString: string) => {
```
Dennull, line 30: Might be worth looking into using date libraries. I think we installed `dayjs` but never use it. Anyways, this is just a tech debt item https://linear.app/rookies-fantasy/issue/ENG-83/consider-using-date-libraries-to-handle-date-operations

client/components/AuthListener.tsx
```diff
@@ -28,9 +30,22 @@ const AuthListener = ({ children }: AuthListenerProps) => {
...
+                    const matchupData = await MatchupController.getUserMatchup(
+                      user.uid,
+                    );
+                    if (matchupData) {
+                      dispatch(setMatchup(matchupData));
+                    }
+                  } catch (error) {
+                    console.error("Error fetching matchup:", error);
+                  }
+                }
```
Dennull, line 48: Sorry if I missed something but why is this needed when we fetch and dispatch on lines 64-67 anyways?

client/app/(protected)/(tabs)/(home)/arena.tsx
Dennull, line 1: I'll be honest I didn't really look through this code that carefully. It looks fine but I suspect since we were rushing we could refactor this to be a bit cleaner as well. I made a ticket to remember to take a look, if it turns out it's all good we can move it to won't do https://linear.app/rookies-fantasy/issue/ENG-85/refactor-arena-component

server/functions/src/index.ts
Dennull: Oh I just read in the description that this is to prevent a race condition. I'm still a bit confused though, could you please explain?

NasihNazeem: The "matching" state is what's been used to stop the race conditions. If user A and user B queue, two different invocations occur for that function and can run concurrently. Without a step in between, the function may run twice, creating two identical matchups. The transaction block ensures that only one of those invocations is valid and the other call gets dismissed.

You're right about this redundant query though. I think I can just store the matched users in a variable that gets initialized outside of the transaction block, and then use that variable. 

I think I'll open up a ticket with regards to optimizing this

server/functions/src/index.ts
NasihNazeem: Yeah it's confusing, I'm down to look it over with you

client/controllers/matchupController.ts
NasihNazeem: This is legacy code, as `matchup.weekStartDate` already gets filled earlier in the controller

client/controllers/matchupController.ts
NasihNazeem: Yes it is, it needs to be done because of the way we store matchup data. We can look into this when restructuring matchup schema

client/components/AuthListener.tsx
NasihNazeem: This is done for the scenario where you are loaded in the app already, and the queue gets updated during an in-app session.

I can remove it if we don't really care for real-time matching. I can create a ticket to maybe track status changes in another way.

server/functions/src/index.ts
NasihNazeem: https://linear.app/rookies-fantasy/issue/ENG-86/optimize-queue-matching-by-eliminating-redundant-firestore-query

client/components/AuthListener.tsx
NasihNazeem: https://linear.app/rookies-fantasy/issue/ENG-87/refactor-tracking-queuestatus-changes-in-authlistener-component

server/functions/src/index.ts
Dennull: Ok sounds good, thanks!

client/components/AuthListener.tsx
Dennull: Oh I see, cause this is the callback that gets passed into the firestore, right?

Dennull (approved):

client/components/AuthListener.tsx
NasihNazeem: I'm not sure I'm understanding but basically I am using this subscriber to listen to updates to the user document, and if there is an update to queueStatus, that's when I fetch for the updated matchup document for that user. There's no subscriber to the matchup document so I would have to do something like this to only check when the queueStatus for that user == matched

PR #34, opened 2025-11-27
Bugfix/ENG-75: "Build your team" button is not pressable again when redirecting back to home page

NasihNazeem:
## Related Issues

Closes ENG-75.

## Summary

This PR fixes a bug where the "Build Your Team" button would only work once and become permanently disabled. The issue was caused by an `isNavigating` state that was set to true and never gets turned false so we never get back to the the router.push() that is inside a `if(!isNavigating){}` statement.

I'm not sure exactly, but I believe the reason we had this is to prevent double-clicks as if I remove the if statement, the only issue I see is being able to click on the button multiple times, causing multiple router pushes. 

## Changes Made

I introduced throttling at the reusable Button component level and this fixed the problem. Throttling removes the need for the `isNavigating` flag as double clicks are not possible with this new function.

## Screenshots

No additional screenshots

## Testing Instructions

Make sure you are using an account with an empty lineup to see the "Build Your Team" button

- [ ] Button works on first press
- [ ] Button works on subsequent presses
- [ ] Rapid double-taps do not cause duplicate navigation
- [ ] Any navigation errors are handled gracefully
- [ ] All buttons throughout the app now are protected from double-taps

No additional notes

### DIFF
--- a/client/components/Button.tsx
+++ b/client/components/Button.tsx
@@ -1,3 +1,4 @@
+import { useCallback, useRef } from "react";
 import { Pressable, Text } from "react-native";
 import Spinner from "./Spinner";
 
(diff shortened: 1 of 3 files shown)


Dennull:
I think originally we added this to prevent the user from spamming the button because they could navigate multiple times. I think I vaguely remember us discussing a debounce or throttle but there might have been a reason why we didn't do it. @JayChuang8 do you remember why?

The only reason I can think of would be if the page take a while to load they can spam it again, but that's so unlikely and is its own problem. So I think I'm gonna approve this but basically wanted to ask if Jay remembers why, if there isn't really a reason this is probably good enough for us rn

client/components/Button.tsx
Dennull, line 1: If you wanted to do it an improvement would be to add an optional debounce prop so that calling components specify the time we want to debounce it for. Maybe `debounceMs` or something

Dennull (approved):

client/components/Button.tsx
NasihNazeem: I went with a throttle for this implementation to specifically stop any calls from happening after the first click. I'll implement a `throttleMs` just like you mentioned so we can configure how long it should be from the calling component

NasihNazeem: Additionally, I can add a `debounceMs` but I'm just not sure about adding it if we aren't using it just yet if that makes sense

PR #35, opened 2025-11-28
Refactor/ENG-74: Import and export headshot url scripts

NasihNazeem:
## Related Issues

Closes ENG-74.

## Summary

We needed a way to transfer headshot URLs as currently there is no simple way of getting the headshots we need since BallDontLie uses different player ids compared to the NBA. This implements export and import scripts that can help transfer urls from development and staging. This really only needs to be run a few times every season.

## Changes Made

- Import and export files generated for headshot urls
- JSON of existing players being transferred.

## Screenshots

No additional screenshots

## Testing Instructions

No additional notes


Dennull:
Just a few questions for my own knowledge, you had to manually compile that json of headshots right? Balldontlie doesn't provide those and there's no easy way to pull it online? And the scripts are just to make it easier to insert into your database? Also, dumb question but you run the script just using node like `node importHeadshotsToFirestore.js`?

Dennull (approved):

NasihNazeem:
> Just a few questions for my own knowledge, you had to manually compile that json of headshots right? Balldontlie doesn't provide those and there's no easy way to pull it online? 

So I was initially going through it manually since BDL doesn't provide them, but I then got AI to generate something like that for me. It wasn't perfect, but it got like 70% of them right and I went through the array of players and tried making sure every player had a headshot, but there's always going to be one or two that have an incorrect headshot when we do it this way. 

> And the scripts are just to make it easier to insert into your database? 

The scripts are to make it easier, but I think I need to revisit the scripts before we deploy again onto production and just make it really easy to deploy the collections without worrying about any migration headache. Maybe I need to document it as well just to make sure it runs smoothly. 

> Also, dumb question but you run the script just using node like `node importHeadshotsToFirestore.js`?

Yes, once I get the appropriate `serviceAccountKey.json` saved into the scripts folder, I just run the script using node

PR #43, opened 2025-12-17
Feature/ENG-35: Player search

JayChuang8:
## Related Issues

Addresses [ENG-35](https://linear.app/rookies-fantasy/issue/ENG-35/search-for-players)

## Summary

- Add search functionality to player list
- Can search by first or last name

## Changes Made

- Instead of [integrating an external search service](https://firebase.google.com/docs/firestore/solutions/search?provider=algolia) to Firestore, a basic "starts with" search is implemented
- For our simple use case of just searching names, this solution will do for now. If our search ever expands to querying through more documents/more complex indexes, we can introduce something like Elasticsearch
- https://code.build/p/firestore-fuzzy-full-text-search-Ut2Smh#idea-2---character-hack
- Right now there is no full name search as we don't have a full name filed in each player document, but we can add this in if we want

## Screenshots

No additional screenshots

## Testing Instructions

No additional notes


client/app/(protected)/(draft)/(teamBuilder)/players.tsx
```diff
@@ -73,6 +73,7 @@ const Players = () => {
       pageSize: PAGE_SIZE,
       pageParam,
       filters: activeFilters ? filters : defaultPlayerFilters,
+      searchTerm: query,
```
Dennull, line 76: We should also add this in the free agents page. There's already a ticket for making that whole table + search bar + filter into one component and reusing that ([ENG-102](https://linear.app/rookies-fantasy/issue/ENG-102/combine-playerstable-with-search-bar-and-filters-component)), but for now we need to manually sync the two pages

client/controllers/nbaPlayersController.ts
Dennull, line 1: This works pretty good! I'd like to add the full name so that the search doesn't break if people try to search by full name. Do you also think we should make it case insensitive? Maybe convert it to lowercase? If we do that I think we have to add a new fullNameLower field because there's no way to transform data at query time

Let me know if you wanna add these changes in this PR or if you want to make a follow up one, I'm cool with whatever

Dennull (changes requested):
There are some failing CI checks to address too

client/app/(protected)/(draft)/(teamBuilder)/players.tsx
JayChuang8: Added search to free agents 👍

client/controllers/nbaPlayersController.ts
JayChuang8: Yeah I added the fullNameLower field in the db now, it supports full name search now 👍

Dennull: Nice!

Dennull (approved):

PR #44, opened 2025-12-21
Feature/ENG-98: Empty arena UI

Dennull:
## Related Issues

Addresses ENG-98

## Summary

Added some UI to cover various states where the user navigates to the matchup page but doesn't have a valid matchup to show yet. These cases include:

1. The user is currently queueing for a match
2. The user is not in queue and is not matched up
3. The user is matched up but for a future week

## Changes Made

Added logic to conditionally check to see which possible matchup state the user is in and render an appropriate UI. I also changed the date selector to always show this current week no matter what, since we're not worrying about displaying past or future matchups yet

## Screenshots

No additional screenshots

## Testing Instructions

No additional notes

### DIFF
--- a/client/app/(protected)/(tabs)/(home)/arena.tsx
+++ b/client/app/(protected)/(tabs)/(home)/arena.tsx
@@ -41,6 +41,7 @@ const Arena = () => {
   const currentDate = selectedDate || availableDates[0] || "";
 
   const dailyMatchup = matchup.dailyMatchups[currentDate];
+  const matchupStartDate = new Date(matchup.weekStartDate);
 
   const awayTeamLogo = teamLogoOptions.find(
     (asset) => asset.url === matchup.away.awayTeamLogo,
@@ -172,7 +173,7 @@ const Arena = () => {
             Searching for an Opponent
           </Text>
           <Text className="pbk-b2 text-center text-modeContrast">
-            {`You're in the queue — finding you a fair matchup. Shouldn't take long.`}
+            {`You're in the queue — finding you a worthy opponent. Shouldn't take long!`}
           </Text>
         </View>
       )}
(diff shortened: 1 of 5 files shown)


Dennull:
Another thing I did here was that I changed the date comparisons in the arena page and date selector to be strictly string comparisons, since we mostly only care about the day for those. It simplifies things because converting to date objects introduces some UTC timezone complexity

There was actually a bug with the date selector that this fixes; previously all of the dates were shifted back by one day since the dates were at midnight UTC, so when we converted it to a date object, since midnight UTC is the previous day in MT the day was being display wrong

client/app/(protected)/(tabs)/(home)/arena.tsx
```diff
@@ -122,170 +152,212 @@ const Arena = () => {
...
+          <Empty color={`rgb(${themes[theme][mode].modeContrast})`} size={32} />
+          <Text className="pbk-h8 text-center text-modeContrast">
+            Check in Next Week!
+          </Text>
+          <Text className="pbk-b2 text-center text-modeContrast">
+            {`You're matched up against an opponent! However, your match is next week. Please come back next week!`}
+          </Text>
+        </View>
+      )}
+      {!matchupEmpty && matchup.weekStartDate === currentWeekDates[0] && (
```
NasihNazeem, line 226: So currently, there is a previous matchup that exists and its showing this state when its actually only meant for those who have `queueStatus === "matched"` along with those checks you have. I think that'll work

NasihNazeem (approved):
Didn't have anything crazy but I did have one suggestion in terms of what state we should be seeing. I think it can cause problems down the line once we have previous matchups. 

I can also understand if you think my suggestion is the wrong one, as technically we should be fetching for present and future matchups only, not previous ones. This check could be handled inside of AuthListener or the controller `getMatchup()`.

client/app/(protected)/(tabs)/(home)/arena.tsx
Dennull: Sorry I don't quite get it, so this needs an extra queueStatus === "matched" check?

NasihNazeem: Okay I looked into it a bit more and I realize this isn't an issue related to this PR so I'm willing to create a hotfix ticket for it as I've found the problem. 

Initially I thought making that extra check at this UI layer would make sense, but I think I take that back and I think it needs to be done within the `AuthListener.tsx` component. Currently we are pulling for a matchup regardless of `queueStatus` being `matched` or not. 

 There are two `getMatchup` calls inside `AuthListener.tsx` for some reason so I'm just going to remove the one that doesn't have that conditional based on `queueStatus`. That should fix it
 
 Let me know if I should set that ticket up @Dennull we can get it sorted quick

Dennull: Yeah I'm fine with getting this fixed in a separate story 👍 I'm gonna merge this then

PR #46, opened 2025-12-23
Feature/ENG-90: Implement augment effects computations

NasihNazeem:
## Related Issues

Closes ENG-90.

## Summary

This PR implements the augment effects computation system that applies stat multiplier bonuses to qualifying players' fantasy points. The system validates augment prerequisites, identifies players who meet all conditions, and calculates augmented fantasy points by applying stat-specific multipliers on top of base fantasy points. Both the cloud function (`updateDailyPlayerData`) and client-side matchup logic now support dynamic fantasy point calculations based on active augments.

## Changes Made

- Added augment computation to `updateDailyPlayerData` scheduled function
    - Validates augment prerequisites
    - Identifies qualifying players
    - Calculates fpts with augment included
- Refactored data structure
    - Remove fields like `playerId`, `firstName`, `lastName`, `gamesPlayed` from stored maps (averageStats and gamelog)
- Refactored `augmentValidation` to become `augmentUtils` to handle applying augment affects along with validation
- Refactor field names throughout screens to have consistent naming on both frontend and backend
- Created an emulator testing script

## Screenshots

No additional screenshots

## Testing Instructions

No additional notes


client/types/augment.ts
```diff
@@ -58,28 +58,3 @@ export type Effect = {
     },
   ];
 };
-
-export type Player = {
```
Dennull, line 62: Was this not being used?

client/utils/augmentUtils.ts
```diff
@@ -203,3 +179,80 @@ const evaluateCondition = (
...
+  stats: GameStats,
+  playerId: string,
+  qualifyingPlayers: Player[] | undefined,
+  augment: Augment | undefined,
+): number => {
+  if (!augment || !augment.isActive || !qualifyingPlayers) {
+    return baseFantasyPoints;
+  }
+
+  const playerQualifies = qualifyingPlayers.some((p) => p.id === playerId);
```
Dennull: Don't we need to do similar logic as what's in `validateAugment` to check if we're able to apply the augment first? As is for example you could have 2/3 qualifying players and the augment would still apply to them, right?

client/utils/augmentUtils.ts
```diff
@@ -203,3 +179,80 @@ const evaluateCondition = (
...
+  if (!augment.effects || augment.effects.length === 0) {
+    return baseFantasyPoints;
+  }
+
+  let totalPoints = baseFantasyPoints;
+
+  augment.effects.forEach((effect) => {
+    effect.statBoosts.forEach((boost) => {
+      const statValue = getStatValue(stats, boost.stat);
+      const baseStatMultiplier = getBaseStatMultiplier(boost.stat);
```
Dennull: Do you think it makes sense to combine these two functions into something like `getBaseStatFantasyPoints` or something? The name could use some work lol, but both use the same switch statement so we could use the same code paths

Dennull (changes requested):

client/types/augment.ts
NasihNazeem: Yeah I'm not sure why it was there originally. We had a Player object already

client/utils/augmentUtils.ts
NasihNazeem: Thanks for catching this. I'll add this check after checking if augment is valid

```
if (qualifyingPlayers.length < augment.playerCount) {
    return baseFantasyPoints;
  }
  ```

File: client/utils/augmentUtils.ts (hunk 3 above)
Comment (NasihNazeem) [reply client/utils/augmentUtils.ts]: Structurally they are similar but I feel like it's best to keep them separate since they serve different functions. For example, `getStatValue` requires different arguments, returns a number similar to the other function, but its only used to get the player stat. The other returns the multiplier, and there are cases where you may need one and not the other.

However, since `getBaseStatMultiplier` is basically a function acting like a map, I created a static record object instead where we just refer to that using the key like `points` or `rebounds`.

```
const STAT_MULTIPLIERS: Record<string, number> = {
  points: 1,
  rebounds: 1,
  assists: 2,
  steals: 3,
  blocks: 3,
  turnovers: -2,
};
```
and the implementation is essentially

```suggestion
      const statValue = getStatValue(stats, boost.stat);
      const baseStatMultiplier = STAT_MULTIPLIERS[boost.stat] ?? 0;
```

client/utils/augmentUtils.ts
Dennull: We discussed this offline, the validation isn't happening properly for the augment on the front end right now. We should evaluate the augment condition and prior to calculating the fantasy points

client/utils/augmentUtils.ts
Dennull: I see, that's good then. Another side comment, do you think it's better if we calculate the fantasy points in here rather than pass it in then? We already have the whole `GameStats` object

Dennull (approved):

PR #47, opened 2025-12-25
Feature/ENG-99: Matchup info card

JayChuang8:
## Related Issues

[ENG-99](https://linear.app/rookies-fantasy/issue/ENG-99/info-card-for-when-matched-but-matchup-is-on-following-week)

## Summary

- Add a matchup info card to display information to the user in different states: letting them know when lineups lock if they are not queueing for a match, and letting them know they need to cancel queueing if they want to make edits to their lineup

## Changes Made

- Add a `getEarliestGameStartTime` cloud function to fetch the earliest game start time, implemented with caching
- Add `MatchupInfoCard` component to show the appropriate message
- Add `useCountdown` custom hook to get the time remaining to a target time
- Only show the card when a user is in a matchup, calculate the earliest game start time based on whether the matchup is underway or next week
- Move balldontlie api calls to common `ballDontLieController.ts`

## Screenshots

image
image

## Testing Instructions

No additional notes

### DIFF
--- a/client/components/MatchupInfoCard.tsx
+++ b/client/components/MatchupInfoCard.tsx
@@ -10,7 +10,7 @@ const EARLIEST_GAME_START_TIME_URL =
   "https://us-central1-rookies-fantasy-development.cloudfunctions.net/getEarliestGameStartTime";
 
 const MatchupInfoCard = () => {
-  const queueStatus: QueueStatus = useAppSelector(
+  const queueStatus: QueueStatus | undefined = useAppSelector(
     (state) => state.user.queueStatus,
   );
   const today = new Date();
--- a/client/components/MatchupInfoCard.tsx
+++ b/client/components/MatchupInfoCard.tsx
@@ -10,9 +10,8 @@ const EARLIEST_GAME_START_TIME_URL =
   "https://us-central1-rookies-fantasy-development.cloudfunctions.net/getEarliestGameStartTime";
 
 const MatchupInfoCard = () => {
-  const queueStatus: QueueStatus | undefined = useAppSelector(
-    (state) => state.user.queueStatus,
-  );
+  const queueStatus: QueueStatus =
+    useAppSelector((state) => state.user.queueStatus) ?? QueueStatus.Idle;
   const today = new Date();
   const formattedDate = today.toLocaleDateString("en-US", {
     month: "short",
--- a/client/components/MatchupInfoCard.tsx
+++ b/client/components/MatchupInfoCard.tsx
@@ -4,14 +4,15 @@ import { useEffect, useState } from "react";
 import { View, Text } from "react-native";
 import { useCountdown } from "@/hooks/useCountdown";
 import { useAppSelector } from "@/state/hooks";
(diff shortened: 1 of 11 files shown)


client/types/user.ts
NasihNazeem, line 1: Thanks for making this enum

server/functions/src/index.ts
NasihNazeem, line 1: I'm not sure if this function has been deployed yet so to test I used the emulator to get it going, but once I did it worked as expected. Right merging this in, we should deploy the function if you haven't already

EDIT: I just saw that you already deployed the function so my build not have been up-to-date. Should be all good to merge

NasihNazeem (approved):

client/hooks/useLineupLock.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  const [earliestStartTime, setEarliestStartTime] = useState<string>();
+
+  useEffect(() => {
+    let cancelled = false;
+
+    const getEarliestGame = async () => {
+      try {
+        const startTime = await fetchEarliestGameStartTime(apiDate);
+        if (!cancelled) {
```
Dennull, line 30: Just a comment, we should implement better cancellation in the future if we ever get there. Idk if there's a ticket for this yet

client/hooks/useLineupLock.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    getEarliestGame();
+
+    return () => {
+      cancelled = true;
+    };
+  }, [apiDate]);
+
+  const isLineupLocked =
+    !!earliestStartTime && new Date() >= new Date(earliestStartTime);
```
Dennull, line 46: Our date handling sucks right now, but I don't want to blow up our tickets to be all about timezones; that's a future tech debt story. However, a band-aid fix for this it to convert `earliestStartTime`that to local time on the client since our API returns it in UTC

client/components/MatchupInfoCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { useAppSelector } from "@/state/hooks";
+import { selectMatchup } from "@/state/slices/matchupSlice";
+import { QueueStatus } from "@/types/user";
+
+const MatchupInfoCard = () => {
+  const queueStatus =
+    useAppSelector((state) => state.user.queueStatus) ?? QueueStatus.Idle;
+  const matchup = useAppSelector(selectMatchup);
+
+  const { isLineupLocked, countdown } = useLineupLock({
```
Dennull, line 12: You think we'd ever want to display the countdown separately? Maybe removing the countdown return from the lineup lock hook is better separation of concerns? But I don't feel that strongly about it, so I'm cool with whatever you wanna go with

Dennull (approved):

client/components/MatchupInfoCard.tsx
JayChuang8: We could always just use the `useCountdown` hook separately if we want just that. I just added it in here since the countdown we need for the lineup lock use case depends on other variables that live in this `useLineupLock` hook, so having the countdown logic in there simplifies things

PR #59, opened 2026-06-07
 Bugfix/ENG-107: Fix archived matchups loading on auth 

NasihNazeem:
## Related Issues

Closes ENG-107.

## Summary

Archived matchups were being unconditionally fetched and loaded into state on every auth state change, causing stale matchups to appear as active.

## Changes Made

- `AuthListener` — Removed the unconditional `getUserMatchup` call on auth load. The `subscribeToUser` real-time listener already handles fetching matchups when `queueStatus === "matched"`, so the duplicate call was redundant and incorrectly loading archived matchups.
- `CLAUDE.md` — Added root-level CLAUDE.md with project structure, conventions, and TypeScript rules for code generation guidance. *This is already in release/0.1 but this one has some small additions*

## Screenshots

No additional screenshots

## Testing Instructions

1. Log in with an account that has archived matchups but is not currently matched.
2. Verify no matchup data is loaded into state.
3. Log in with an account where queueStatus === "matched" and confirm the active matchup loads correctly.

No additional notes


client/components/AuthListener.tsx
```diff
@@ -59,11 +59,6 @@ const AuthListener = ({ children }: AuthListenerProps) => {
             );
             dispatch(setTeam(teamData));
           }
-
-          const matchupData = await MatchupController.getUserMatchup(user.uid);
```
Dennull, line 65: So if someone is trying to view a past matchup, they don't be able to anymore? I think we said we'd deal with past history later, do you remember if that's right?

NasihNazeem: The intention on this fetch is to get current matchups to show in the arena page when loading up the data for the current user. 

Let's assume a user doesn't have a current matchup. In that case, the arena page should default to showing that screen that says "you don't have a matchup" and not automatically showing a previous matchup. If they want to look at a previous matchup we should make it accessible through the arena page itself by fetching for that data when they request it. So you're right in saying we said we would deal with it later

I hope that makes some sense. Let me know if you agree with that

JayChuang8 (approved):

PR #60, opened 2026-06-07
Bugfix/ENG-119: Return 200 with null earliestGameStart when no games found

JayChuang8:
## Related Issues

Closes ENG-119

## Summary

The `getEarliestGameStartTime` cloud function was returning a `404` response when no games were scheduled for the requested date (e.g. off-season dates). This caused the client to attempt to parse an HTML "Not Found" string as JSON, resulting in a `SyntaxError: JSON Parse error: Unexpected character: N` crash.

A 404 is semantically incorrect here — the endpoint exists and the request is valid; there simply are no games on that date. This is an expected, non-error state.

## Changes Made

- **`server/functions/src/index.ts`**: Changed the no-games branch in `getEarliestGameStartTime` from `res.status(404).send("No games found")` to `res.status(200).json({ earliestGameStart: null })`, so the client always receives valid JSON.
- **`client/controllers/ballDontLieController.ts`**: Removed the workaround `response.ok` guard and extra logging that were added to handle the erroneous 404, since the server now always returns a parseable response.

## Screenshots

No additional screenshots

## Testing Instructions

1. Call `fetchEarliestGameStartTime` with an off-season date (e.g. `2026-06-07`).
2. Confirm no `SyntaxError` is thrown and the function returns `undefined`.
3. Call with an in-season date that has games and confirm the correct start time is still returned.

No additional notes

### DIFF
--- a/server/functions/src/index.ts
+++ b/server/functions/src/index.ts
@@ -1363,7 +1363,7 @@ export const getEarliestGameStartTime = functions.https.onRequest(
       });
 
       if (!data || data.length === 0) {
-        res.status(404).send("No games found");
+        res.status(200).json({ earliestGameStart: null });
         return;
       }
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,9 @@
+# macOS
+.DS_Store
+
+# IDE
+.idea/
+
+# Expo generated native folders (use `npx expo prebuild` to regenerate)
+client/ios/
+client/android/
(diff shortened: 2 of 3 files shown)


server/functions/src/index.ts
Dennull, line 1: If the front end receives a null start time, does it know how to handle that? Otherwise we might be displaying null by accident now instead of getting a 404

JayChuang8: I changed the message to show "No games today" now

Dennull (approved):

